PR #32, opened 2018-05-18
Location bulk upload

arvindyadav108:
Issue # SB-3109 - Bulk upload scheduler fixes.

### DIFF
--- a/actors/common/src/main/java/org/sunbird/learner/actors/SchedularActor.java
+++ b/actors/common/src/main/java/org/sunbird/learner/actors/SchedularActor.java
@@ -9,6 +9,7 @@ import org.sunbird.cassandra.CassandraOperation;
 import org.sunbird.common.models.util.ActorOperations;
 import org.sunbird.common.models.util.BulkUploadActorOperation;
 import org.sunbird.common.models.util.JsonKey;
+import org.sunbird.common.models.util.LoggerEnum;
 import org.sunbird.common.models.util.ProjectLogger;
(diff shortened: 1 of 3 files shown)


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/LocationBulkUploadBackGroundJobActor.java
```diff
@@ -77,6 +77,8 @@ public void onReceive(Request request) throws Throwable {
   private void bulkLocationUpload(Request request) throws IOException {
 
     String processId = (String) request.get(JsonKey.PROCESS_ID);
+    ProjectLogger.log(
+        "LocationBulkUpload Background job started for Id : " + processId, LoggerEnum.INFO);
```
bvinayakumar: Change log statement as follows:

"LocationBulkUploadBackGroundJobActor: bulkLocationUpload called with process ID " + processId

actors/common/src/main/java/org/sunbird/learner/actors/SchedularActor.java
```diff
@@ -69,7 +70,8 @@ private void schedule(Request request) {
             "calling bulkUploadBackGroundJobActor for processId from schedular actor "
```
bvinayakumar: Change log statement as:

"SchedularActor: scheduleBulkUpload called with processId " + {0} + " and type " + {1}

PR #45, opened 2018-05-29
Org refactoring

arvindyadav108:
SB-3955 : Org bulk upload not working.


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BulkUploadManagementActor.java
```diff
@@ -387,7 +387,11 @@ private void uploadCsvToDB(
         String[] columnArr = dataList.get(0);
         columnArr = trimColumnAttributes(columnArr);
         Map<String, Object> dataMap = null;
-        String channel = Util.getChannel(rootOrgId);
+        String channel = null;
+        // channel is required only in case of the user type bulk upload.
+        if (!StringUtils.isBlank(objectType) && objectType.equalsIgnoreCase(JsonKey.USER)) {
```
bvinayakumar: Change to isNotBlank.

PR #54, opened 2018-05-31
Issue #SB-3374 feat: Sunbird-OpenSaber-Integration

JaikumarSRajan:
UserManagementActor updated to call UserExtension flow during User Creation


actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -1691,6 +1693,17 @@ private void createUser(Request actorMessage) {
     response.put(JsonKey.ACCESSTOKEN, accessToken);
     sender().tell(response, self());
 
+    /*
+     * Create User Entity in Registry
+     */
+    if(ProjectUtil.getConfigValue("sunbird_open_saber_bridge_enable").equalsIgnoreCase("true")){
```
bvinayakumar: Format code.

bvinayakumar: Can we change condition as "true".equalsIgnoreCase(...)?

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -1691,6 +1693,17 @@ private void createUser(Request actorMessage) {
...
+		 String firstName = (String) actorMessage.getRequest().get(JsonKey.FIRST_NAME);
+		 Map<String,Object> extensionMap = new HashMap<String,Object>();
```
bvinayakumar: Instead of creating a new hashmap, can we pass the received create user request map as it is?

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -685,6 +686,15 @@ private void getUserProfile(Request actorMessage) {
       result.remove(JsonKey.MISSING_FIELDS);
       result.remove(JsonKey.COMPLETENESS);
     }
+    /*
```
bvinayakumar: As discussed today read will be from ES. So, read API call not needed in getUserProfile.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
JaikumarSRajan: Code formatting done using git pre-hook

JaikumarSRajan: Review comment incorporated in code

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
JaikumarSRajan: Review Comment Incorporated in Code

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
JaikumarSRajan: Read API Call is removed from getUserProfile method

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -673,6 +673,28 @@ public static String getUserNamebyUserId(String userId) {
     }
     return null;
   }
+
+  /**
+   * This method will provide user details map based on user id if user not found then it will
+   * return null.
+   *
+   * @param userId String
```
bvinayakumar: Javadoc comment.

PR #112, opened 2018-08-17
add check of same org

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/skill/UserSkillManagementActor.java
```diff
@@ -556,4 +563,42 @@ private void addTelemetry(String userId, Request request) {
...
+            userDbInfo.getKeySpace(), userDbInfo.getTableName(), requestedUserId);
+    List<Map<String, Object>> endoresedList =
+        (List<Map<String, Object>>) endorsedUserResponse.get(JsonKey.RESPONSE);
+    List<Map<String, Object>> requestedUserList =
+        (List<Map<String, Object>>) requestedUserResponse.get(JsonKey.RESPONSE);
+
+    // check whether both userid exist or not if not throw exception
+    if (endoresedList.isEmpty() || requestedUserList.isEmpty()) {
+      // generate context and params here ...
+      throw new ProjectCommonException(
```
bvinayakumar: Use invalidParameterValue.

PR #117, opened 2018-08-20
fix issues with multiple count and timestamp

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/skill/dao/impl/UserSkillDaoImpl.java
```diff
@@ -55,16 +59,12 @@ public Skill read(String id) {
   @Override
   public void update(Skill skill) {
     ObjectMapper objectMapper = new ObjectMapper();
-    objectMapper.configure(SerializationFeature.WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS, false);
-    objectMapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, true);
     HashMap<String, Object> map =
         (HashMap<String, Object>) objectMapper.convertValue(skill, Map.class);
     if (map.containsKey("createdOn")) {
```
manzarul: use json key "createdOn" instead of hard coding  it.

actors/common/src/main/java/org/sunbird/learner/actors/skill/dao/impl/UserSkillDaoImpl.java
```diff
@@ -55,16 +59,12 @@ public Skill read(String id) {
...
       map.remove("createdOn");
     }
-    if (map.containsKey("lastUpdatedOn")) {
-      map.remove("lastUpdatedOn");
-    }
+    map.put("lastUpdatedOn", new Timestamp(Calendar.getInstance().getTime().getTime()));
```
manzarul: lastUpdatedOn using  JsonKey class to put this constant.

PR #120, opened 2018-08-21
Issues #SB-5530 fix: changes for course batch

kirtisagarksj:


actors/common/src/main/java/org/sunbird/learner/actors/CourseBatchManagementActor.java
```diff
@@ -640,7 +631,20 @@ private void updateCourseBatch(Request actorMessage) {
...
         }
-
+        // kirti
+        if (dbBatchStartDate.compareTo(todate) == 0) {
+          throw new ProjectCommonException(
+              ResponseCode.invalidBatchStartDateError.getErrorCode(),
+              ResponseCode.invalidBatchStartDateError.getErrorMessage(),
+              ResponseCode.CLIENT_ERROR.getResponseCode());
+        } else {
+          if (requestedStartDate.before(todaydate)) {
```
bvinayakumar: This check is not required due to above comment.

actors/common/src/main/java/org/sunbird/learner/actors/CourseBatchManagementActor.java
```diff
@@ -640,7 +631,20 @@ private void updateCourseBatch(Request actorMessage) {
                 ResponseCode.CLIENT_ERROR.getResponseCode());
           }
         }
-
+        // kirti
+        if (dbBatchStartDate.compareTo(todate) == 0) {
```
bvinayakumar: Use equals instead of compareTo.

actors/common/src/main/java/org/sunbird/learner/actors/CourseBatchManagementActor.java
```diff
@@ -640,7 +631,20 @@ private void updateCourseBatch(Request actorMessage) {
                 ResponseCode.CLIENT_ERROR.getResponseCode());
           }
         }
-
+        // kirti
```
bvinayakumar: Remove comment with your name.

PR #122, opened 2018-08-24
Issue #SB-5661 feat:saving external ids in original case not in lower…

amit-tarento:
…case


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BulkUploadBackGroundJobActor.java
```diff
@@ -1429,6 +1429,8 @@ private void updateUser(Map<String, Object> userMap, Map<String, Object> userDbR
 
   private void validateExternalIds(Map<String, Object> userMap, String operation) {
     if (CollectionUtils.isNotEmpty((List<Map<String, String>>) userMap.get(JsonKey.EXTERNAL_IDS))) {
```
bvinayakumar: Util.copyAndConvertExternalIdsToLower

PR #142, opened 2018-09-10
Issue #SB-0000 fix: change code according to controller

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/NotesManagementActor.java
```diff
@@ -41,18 +41,24 @@ public void onReceive(Request requsst) throws Throwable {
...
-      updateNote(requsst);
-    } else if (requsst.getOperation().equalsIgnoreCase(ActorOperations.SEARCH_NOTE.getValue())) {
-      searchNote(requsst);
-    } else if (requsst.getOperation().equalsIgnoreCase(ActorOperations.GET_NOTE.getValue())) {
-      getNote(requsst);
-    } else if (requsst.getOperation().equalsIgnoreCase(ActorOperations.DELETE_NOTE.getValue())) {
-      deleteNote(requsst);
-    } else {
-      onReceiveUnsupportedOperation(requsst.getOperation());
+    switch (requsst.getOperation()) {
```
bvinayakumar: requsst -> request

actors/common/src/main/java/org/sunbird/learner/actors/NotesManagementActor.java
```diff
@@ -154,14 +160,13 @@ private void updateNote(Request actorMessage) {
...
       req.remove(JsonKey.COURSE_ID);
       req.remove(JsonKey.CONTENT_ID);
       req.remove(JsonKey.IS_DELETED);
       req.put(JsonKey.ID, noteId);
-      String updatedBy = (String) actorMessage.getRequest().get(JsonKey.REQUESTED_BY);
-      req.put(JsonKey.UPDATED_BY, updatedBy);
+      req.put(JsonKey.UPDATED_BY, userId);
       req.put(JsonKey.UPDATED_DATE, ProjectUtil.getFormattedDate());
 
       Response result =
```
bvinayakumar: Since the map contains context. It may cause error when updating record. This may need to be removed from map.

actors/common/src/main/java/org/sunbird/learner/actors/NotesManagementActor.java
```diff
@@ -335,8 +334,7 @@ private void deleteNote(Request actorMessage) {
       Map<String, Object> req = new HashMap<>();
       req.put(JsonKey.ID, noteId);
       req.put(JsonKey.IS_DELETED, true);
-      String updatedBy = (String) actorMessage.getRequest().get(JsonKey.REQUESTED_BY);
-      req.put(JsonKey.UPDATED_BY, updatedBy);
+      req.put(JsonKey.UPDATED_BY, userId);
```
bvinayakumar: Since the map contains context. It may cause error when updating record. This may need to be removed from map. Test this scenario.

PR #160, opened 2018-09-27
Issue #SB-7061 feat: User Profile Visibility - Global Settings

JaikumarSRajan:


actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -1463,24 +1467,86 @@ public static void checkProfileCompleteness(Map<String, Object> userMap) {
...
+    Map<String, String> userProfileVisibilityMap =
         (Map<String, String>) userMap.get(JsonKey.PROFILE_VISIBILITY);
-    if (null != profileVisibility && !profileVisibility.isEmpty()) {
-      Map<String, Object> profileVisibilityMap = new HashMap<>();
-      for (String field : profileVisibility.keySet()) {
-        profileVisibilityMap.put(field, userMap.get(field));
+    Map<String, String> completeProfileVisibilityMap =
+        getCompleteProfileVisibilityMap(userProfileVisibilityMap, actorRef);
+
+    if (null != completeProfileVisibilityMap && !completeProfileVisibilityMap.isEmpty()) {
```
bvinayakumar: For brevity MapUtils.isEmpty can be used.

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -1463,24 +1467,86 @@ public static void checkProfileCompleteness(Map<String, Object> userMap) {
...
+      }
+    }
+    return completeProfileVisibilityMap;
+  }
+
+  public static Map<String, String> getCompleteProfileVisibilityMap(
+      Map<String, String> userProfileVisibilityMap, ActorRef actorRef) {
+    String defaultProfileVisibility =
+        ProjectUtil.getConfigValue(JsonKey.SUNBIRD_DEFAULT_USER_PROFILE_VISIBILITY);
+    if (!(defaultProfileVisibility.equals(JsonKey.PUBLIC)
```
bvinayakumar: Use ignore case in comparison.

Use JsonKey.PUBLIC.equals as that would avoid null pointer issue.

bvinayakumar: Please have default value as 'public' so that other adopters are not affected.

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -1463,24 +1467,86 @@ public static void checkProfileCompleteness(Map<String, Object> userMap) {
...
+        || defaultProfileVisibility.equals(JsonKey.PRIVATE))) {
+      ProjectLogger.log(
+          "Util:getCompleteProfileVisibilityMap: Invalid configuration - "
+              + defaultProfileVisibility
+              + " - for default profile visibility(public/private)",
+          LoggerEnum.ERROR.name());
+      ProjectCommonException.throwServerErrorException(
```
bvinayakumar: We may not need to throw this error with a valid default.

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -1918,6 +1984,124 @@ public static String getCustodianChannel(Map<String, Object> userMap, ActorRef a
...
   }
+
+  public static void removeEternalFields(
+      String visibilityType, List<String> fieldList, ActorRef actorRef) {
+    Config userProfileConfig = getUserProfileConfig(actorRef);
+    List<String> eternalPublicFields =
+        userProfileConfig.getStringList(JsonKey.ETERNAL_PUBLIC_FIELDS);
+    List<String> eternalPrivateFields =
+        userProfileConfig.getStringList(JsonKey.ETERNAL_PRIVATE_FIELDS);
+    if (JsonKey.PRIVATE.equals(visibilityType)) {
```
bvinayakumar: Ignore case.

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -1918,6 +1984,124 @@ public static String getCustodianChannel(Map<String, Object> userMap, ActorRef a
...
+    SystemSetting userProfileConfigSetting =
+        getSystemSettingById(JsonKey.USER_PROFILE_CONFIG, actorRef);
+    String userProfileConfigString = userProfileConfigSetting.getValue();
+    Config userProfileConfig = ConfigUtil.parseJSONString(userProfileConfigString);
+    validateUserProfileConfig(userProfileConfig);
+    return userProfileConfig;
+  }
+
+  private static void validateUserProfileConfig(Config userProfileConfig) {
+    if (CollectionUtils.isEmpty(userProfileConfig.getStringList(JsonKey.USER_DATA_FIELDS))) {
```
bvinayakumar: Not all adopters may configure userProfileConfig. Can we have a fallback so existing adopters are not affected.

bvinayakumar: Let us discuss this later today.

PR #164, opened 2018-09-28
Issue #SB-0000 add validation for userID

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/NotesManagementActor.java
```diff
@@ -410,11 +410,16 @@ private Boolean noteIdExists(String noteId) {
   private Boolean validateUserForNoteUpdation(String userId, String noteId) {
     Boolean result = false;
     Map<String, Object> noteData = getNoteRecordById(noteId);
-    if ((null != noteData && !noteData.isEmpty())
-        && !StringUtils.isBlank(userId)
-        && userId.equalsIgnoreCase((String) noteData.get(JsonKey.USER_ID))) {
+    if (noteData == null || noteData.size() == 0) return result;
```
bvinayakumar: Use MapUtils.isEmpty

actors/common/src/main/java/org/sunbird/learner/actors/NotesManagementActor.java
```diff
@@ -410,11 +410,16 @@ private Boolean noteIdExists(String noteId) {
...
+    if ((null != noteData && !noteData.isEmpty()) && !StringUtils.isBlank(userId)) {
```
bvinayakumar: Check on noteData is not needed.

PR #187, opened 2018-10-16
Unenroll api actor test

sudhirgiri2911:
user controller - block and unblock


actors/common/src/main/java/org/sunbird/learner/actors/geolocation/GeoLocationManagementActor.java
```diff
@@ -23,7 +22,10 @@
 import org.sunbird.common.request.Request;
 import org.sunbird.common.responsecode.ResponseCode;
 import org.sunbird.helper.ServiceFactory;
+import org.sunbird.learner.actors.geolocation.dao.GeolocationDao;
```
bvinayakumar: Why is this PR having changes relating to GeoLocation? Please create a proper PR for review.

PR #189, opened 2018-10-17
Issue #SB-0000 feat: User Refactoring Code

sudhirgiri2911:
User Refactoring Code

PR #196, opened 2018-10-23
UserManagement actor refactor - part 1

amolnin:
Seperated the actor's inline with controller's division.

Implemented the fixes for UserProfileActor -> like refactoring and using
UserDao.
Introduced UserBaseActor - sharing common functions/objects required by
all the actors.

Please only review userProfileActor, other actors code is just
seperated/moved to relevant classes, refactoring and changes like
moving/using relevant Data access object is yet to be done.

UserManagementActor will contain only create/update/search - which is
being refactored by Amit.


bvinayakumar:
Closing this PR as it contains files from Sudhir's PR (#189) and with missing rework done.

PR #201, opened 2018-10-25
Issue #SB-4221 fix: handle exception

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BaseBulkUploadBackgroundJobActor.java
```diff
@@ -59,8 +61,8 @@ public void handleBulkUploadBackground(Request request, Function function) {
     }
 
     Integer status = bulkUploadProcess.getStatus();
-    if (!(status == (ProjectUtil.BulkProcessStatus.COMPLETED.getValue())
-        || status == (ProjectUtil.BulkProcessStatus.INTERRUPT.getValue()))) {
+    if (!(status.equals(ProjectUtil.BulkProcessStatus.COMPLETED.getValue())
```
amolnin: reverse the condition - constant first, and status as argument.

PR #202, opened 2018-10-25
Issue #SB-4221 Fix: Change required field for bulk upload

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/OrgBulkUploadActor.java
```diff
@@ -64,11 +56,44 @@ private void processOrgBulkUpload(
     if (null != req.get(JsonKey.FILE)) {
       fileByteArray = (byte[]) req.get(JsonKey.FILE);
     }
-    Integer recordCount = validateAndParseRecords(fileByteArray, processId, new HashMap<>());
+    HashMap<String, Object> additionalInfo = new HashMap<>();
+    Map<String, Object> user = getUser((String) req.get(JsonKey.CREATED_BY));
+    if (user != null) {
```
manzarul: what will happen if either user or org will null.

PR #207, opened 2018-10-29
Issue #SB-8012 feat: 

sudhirgiri2911:
Removed the return statement after exception is thrown


actors/common/src/main/java/org/sunbird/learner/actors/coursebatch/CourseEnrollmentActor.java
```diff
@@ -278,12 +255,7 @@ private void validateCourseBatch(
 
   private void verifyRequestedByAndThrowErrorIfNotMatch(String userId, String requestedBy) {
     if (!(userId.equals(requestedBy))) {
-      sender()
-          .tell(
-              createExceptionObject(
-                  ResponseCode.UNAUTHORIZED, ResponseCode.UNAUTHORIZED.getResponseCode()),
-              self());
-      return;
+      createExceptionObject(ResponseCode.UNAUTHORIZED, ResponseCode.UNAUTHORIZED.getResponseCode());
```
bvinayakumar: createExceptionObject only creates the exception object. Instead define and use ProjectCommonException.throwUnauthorizedErrorException.

PR #227, opened 2018-11-12
Bulk upload status

amolnin:


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BulkUploadManagementActor.java
```diff
@@ -141,17 +142,12 @@ private void getBulkUploadDownloadStatusLink(Request actorMessage) {
...
+      } catch (IOException e) {
         ProjectCommonException.throwClientErrorException(
             ResponseCode.errorGenerateDownloadLink, null);
       }
     } else {
-      throw new ProjectCommonException(
-          ResponseCode.invalidProcessId.getErrorCode(),
-          ResponseCode.invalidProcessId.getErrorMessage(),
-          ResponseCode.RESOURCE_NOT_FOUND.getResponseCode());
+      ProjectUtil.createResourceNotFoundException();
```
bvinayakumar: throwResourceNotFoundException needs to be part of ProjectCommonException.

PR #230, opened 2018-11-12
Bulk upload User and Org checks

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BaseBulkUploadActor.java
```diff
@@ -342,14 +342,20 @@ protected void performBatchUpdate(List<BulkUploadProcessTask> records) {
   protected void validateFileHeaderFields(
       Map<String, Object> req, String[] bulkAllowedFields, Boolean allFieldsMandatory)
       throws IOException {
-    validateFileHeaderFields(req, bulkAllowedFields, allFieldsMandatory, false);
+    validateFileHeaderFields(req, bulkAllowedFields, allFieldsMandatory, false,null,null);
```
bvinayakumar: Format code changes.

PR #234, opened 2018-11-13
Masked data for e-mail and phone number

amolnin:
available in output of statusDownloadLink.

In UserBulkUpload, getOrg was returning empty map, so corrected the
condition, so that invalid org id is handled correctly.

### DIFF
--- a/actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BaseBulkUploadBackgroundJobActor.java
+++ b/actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BaseBulkUploadBackgroundJobActor.java
@@ -278,6 +278,7 @@ public abstract class BaseBulkUploadBackgroundJobActor extends BaseBulkUploadAct
         .stream()
         .forEach(
             map -> {
+              preProcessResult(map);
               String[] nextLine = new String[headerRows.size()];
               String errMsg = (String) map.get(JsonKey.ERROR_MSG);
               int i = 0;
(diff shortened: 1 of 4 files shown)


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/UserBulkUploadBackgroundJobActor.java
```diff
@@ -204,9 +212,15 @@ private void callUpdateUser(User user, BulkUploadProcessTask task)
...
     }
     return null;
   }
+  
+  @Override
+  public void convertDataForUserConsumption(Map<String, Object> map) {
+   UserUtility.decryptUserData(map);
+   Util.addMaskEmailAndPhone(map);
+  }
 }
```
bvinayakumar: Why sunbird-utils is being updated in this PR?

actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BaseBulkUploadBackgroundJobActor.java
```diff
@@ -293,4 +294,8 @@ private void addResults(
               csvWriter.writeNext(nextLine);
             });
   }
+
+  public void convertDataForUserConsumption(Map<String, Object> map) {
```
bvinayakumar: convertDataForUserConsumption -> preProcessResult(result)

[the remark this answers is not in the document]
bvinayakumar: Resolve codacy issues.

PR #244, opened 2018-11-15
Org email add

sudhirgiri2911:
Added Validation for email in create and update user


actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -338,6 +339,14 @@ private void createOrg(Request actorMessage) {
         request.put(JsonKey.ROOT_ORG_ID, JsonKey.DEFAULT_ROOT_ORG_ID);
       }
 
+      if (request.containsKey(JsonKey.EMAIL)
+          && !EmailValidator.isEmailValid((String) request.get(JsonKey.EMAIL))) {
+        throw new ProjectCommonException(
```
bvinayakumar: ProjectCommonException.throwClientError(ResponseCode.emailFormatError)

PR #249, opened 2018-11-16
Org email add

sudhirgiri2911:
UnitTest for actor


service/src/test/java/org/sunbird/learner/actors/OrganisationManagementActorTest.java
```diff
@@ -12,6 +12,7 @@
 import java.util.HashMap;
```
bvinayakumar: All these unit tests need to be rewritten as there is lot of code duplication. For example, refer UserSkillManagementActorTest. Also, ensure that these tests pass without Cassandra, ES and Keycloak running.

PR #254, opened 2018-11-18
Externalid support user bulkupload

amit-tarento:


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/UserBulkUploadBackgroundJobActor.java
```diff
@@ -125,7 +125,8 @@ private void processUser(BulkUploadProcessTask task, String organisationId, Stri
       try {
         String roles = (String) userMap.get(JsonKey.ROLES);
         if (roles != null) {
-          userMap.put(JsonKey.ROLES, Arrays.asList(roles.split("\\\\s*,\\\\s*")));
+          String[] roleList = roles.split(",");
```
manzarul: in Roles empty space need to be trim as well. in previous code u are doing.

userMap.put(JsonKey.ROLES, Arrays.asList(roles.split("\\\\s*,\\\\s*")));

actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BulkUploadManagementActor.java
```diff
@@ -214,6 +215,28 @@ private void getUploadStatus(Request actorMessage) {
                       addTaskDataToList(failureList, x.getFailureResult());
                     }
                   });
+          if (!(JsonKey.USER.equalsIgnoreCase(objectType))) {
```
manzarul: JsonKey.USER.equalsIgnoreCase(objectType) - why this check is not for user. it should be for user only. Please check

PR #257, opened 2018-11-21
Issue #SB-6729 feat: Course notifications bugs fixes

iostream04:


actors/common/src/main/java/org/sunbird/learner/actors/coursebatch/CourseBatchNotificationActor.java
```diff
@@ -43,6 +44,11 @@
   private DecryptionService decryptionService =
       org.sunbird.common.models.util.datasecurity.impl.ServiceFactory.getDecryptionServiceInstance(
           null);
+  private static String courseBatchNotificationSignature =
```
bvinayakumar: Why JsonKey.SUNBIRD_COURSE_BATCH_NOTIFICATION_SIGNATURE is required? How are the existing email template signatures working without this? Where is the default value defined for this variable?

iostream04: Existing email template are using ORG name . Product Manager want it configurable. 
Default value is sunbird, it is added to env variable tracking google doc. and also informed in devops channel

bvinayakumar: Is this configurable per tenant?

iostream04: Yes

PR #259, opened 2018-11-22
User skill management actor test

kirtisagarksj:
User skill management actor test


service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
```diff
@@ -315,20 +322,28 @@ private Response createGetUserFailureResponse() {
...
     filters.put(JsonKey.USER_ID, userId);
     esDtoMap.put(JsonKey.FILTERS, filters);
     List<String> fields = new ArrayList<>();
     fields.add(JsonKey.SKILLS);
     esDtoMap.put(JsonKey.FIELDS, fields);
     PowerMockito.mockStatic(ElasticSearchUtil.class);
+
+    if (!isFailure) {
+      response = createGetSkillResponse();
+    } else response = new HashMap<>();
```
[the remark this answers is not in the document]
amolnin: response = isFailure?Collections.emptyMap():createGetSkillResponse

service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
```diff
@@ -315,20 +322,28 @@ private Response createGetUserFailureResponse() {
...
+
     when(ElasticSearchUtil.complexSearch(
             ElasticSearchUtil.createSearchDTO(esDtoMap),
             ProjectUtil.EsIndex.sunbird.getIndexName(),
             ProjectUtil.EsType.user.getTypeName()))
-        .thenReturn(createGetSkillResponse());
+        .thenReturn(response);
```
amolnin: Within when (mocking), use matchers - not exact values. 
May not work sometimes - because you are expecting exact objects, rather than the matchers.

service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
```diff
@@ -48,7 +48,7 @@
   private static final String ENDORSED_USER_ID = "someEndorsedUserId";
   private static final String ENDORSE_SKILL_NAME = "C";
   private static List<String> skillsList = new ArrayList<>();
-  private static FiniteDuration duration = duration("10 second");
+  private static FiniteDuration duration = duration("10000 second");
```
amolnin: What time it generally takes -> we cannot wait for 10000 second, that is too large interval - try out smaller intervals, and see if that works?

service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
```diff
@@ -206,7 +212,8 @@ public void testAddSkillEndorsementSuccess() {
     ActorRef subject = system.actorOf(props);
     mockCasandraRequestForEndorsement();
     subject.tell(
-        createSkillEndorsementRequest(USER_ID, ENDORSED_USER_ID, ENDORSE_SKILL_NAME), probe.getRef());
+        createSkillEndorsementRequest(USER_ID, ENDORSED_USER_ID, ENDORSE_SKILL_NAME),
```
amolnin: Why expecting client_eror - in success test-case?

service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
```diff
@@ -180,7 +183,8 @@ public void testAddSkillEndorsementSuccessWithInvalidEndorsedUserId() {
 
     mockCasandraRequestForGetUser(true);
     subject.tell(
-        createSkillEndorsementRequest(USER_ID, ENDORSED_USER_ID, ENDORSE_SKILL_NAME), probe.getRef());
+        createSkillEndorsementRequest(USER_ID, ENDORSED_USER_ID, ENDORSE_SKILL_NAME),
```
amolnin: This test-case is failure -> wrongly named to success scenario. Finally we are expecting a client_error

amolnin: Casandra is incorrectly spelled in all the methods.

service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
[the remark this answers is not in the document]
amolnin: Intead of simply true/false boolean based, mocking can be done based on arguments to be passed. don't mock from each method to get desired results, rather declare upfront:

like - if argument value matches this then return response, else throw error.

service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
```diff
@@ -39,26 +41,26 @@
...
-  private static final String USER_ID = "someUserId";
+  private CassandraOperation cassandraOperation;
+  private static final String USER_ID = "userId";
   private static final String ROOT_ORG_ID = "someRootOrgId";
   private static final String ENDORSED_USER_ID = "someEndorsedUserId";
-  private static final String ENDORSE_SKILL_NAME = "C";
-  private static List<String> skillsList = new ArrayList<>();
-  private static FiniteDuration duration = duration("10 second");
+  private static final String ENDORSED_SKILL_NAME = "someEndorsedSkillName";
+  private List<String> skillsList = new ArrayList<>();
```
amolnin: skillsList is never populated why is it used.

service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
```diff
@@ -39,26 +41,26 @@
...
+  private FiniteDuration duration = duration("10 second");
```
amolnin: This should remain static final? why have we changed this to class member.

PR #265, opened 2018-11-27
Issue #SB-8793 feat:checkoing email Uniqueness

amit-tarento:


actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -464,6 +467,29 @@ private void createOrg(Request actorMessage) {
     }
   }
 
+  @SuppressWarnings("unchecked")
+  private boolean checkEmailUniqueness(Map<String, Object> orgMap, String operationType) {
```
bvinayakumar: Add this method in OrgService.

validateEmailUniqueness

bvinayakumar: operationType -> isOrgCreate (boolean)

actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -464,6 +467,29 @@ private void createOrg(Request actorMessage) {
...
+    Response result =
+        cassandraOperation.getRecordsByIndexedProperty(
+            orgDbInfo.getKeySpace(),
+            orgDbInfo.getTableName(),
+            JsonKey.EMAIL,
```
bvinayakumar: Is there an index on organisation.email ?

PR #267, opened 2018-11-28
Changes to add Process Id, Status, Message, Object Type to output of

amolnin:
status download link


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BulkUploadManagementActor.java
```diff
@@ -140,7 +140,12 @@ private void getBulkUploadDownloadStatusLink(Request actorMessage) {
                 cloudStorageData.getFileName());
         Response response = new Response();
         response.setResponseCode(ResponseCode.OK);
-        response.getResult().put(JsonKey.SIGNED_URL, signedUrl);
+        Map<String,Object> resultMap  = response.getResult();
+        resultMap.put(JsonKey.SIGNED_URL, signedUrl);
+        resultMap.put(JsonKey.OBJECT_TYPE, bulkUploadProcess.getObjectType());
+        resultMap.put(JsonKey.PROCESS_ID, bulkUploadProcess.getId());
+        resultMap.put(JsonKey.STATUS, bulkUploadProcess.getStatus());
+        updateResponseStatus(resultMap);
```
bvinayakumar: Add "message" as well.

PR #273, opened 2018-11-30
User profile actor test2

kirtisagarksj:
User profile actor test2


actors/user/src/test/java/org/sunbird/user/UserProfileActorTest.java
```diff
@@ -0,0 +1,135 @@
...
+  Util.class,
+})
+@PowerMockIgnore({"javax.management.*"})
+public class UserProfileActorTest {
+
+  private ActorSystem system = ActorSystem.create("system");
+
+  private final Props props = Props.create(UserProfileActor.class);
+  private CassandraOperationImpl cassandraOperation;
+  private SearchDTO searchDTO;
```
amolnin: No need to delcare at class level

actors/user/src/test/java/org/sunbird/user/UserProfileActorTest.java
```diff
@@ -0,0 +1,135 @@
...
+  private String userId = "USER-ID";
```
amolnin: private static final String USER_ID =

actors/user/src/test/java/org/sunbird/user/UserProfileActorTest.java
```diff
@@ -0,0 +1,135 @@
...
+    ActorRef subject = system.actorOf(props);
+
+    when(ElasticSearchUtil.getDataByIdentifier(
+            ProjectUtil.EsIndex.sunbird.getIndexName(),
+            ProjectUtil.EsType.user.getTypeName(),
+            userId))
+        .thenReturn(createGetResponse(true));
+    Request reqObj = new Request();
+    reqObj.setOperation(ActorOperations.PROFILE_VISIBILITY.getValue());
+    reqObj.put(JsonKey.USER_ID, userId);
```
amolnin: above 3 lines are repeated in all test-cases 
variable part is actoroperation.

so create a funciton which prepares request object -> takes actoroperation as parameter

actors/user/src/test/java/org/sunbird/user/UserProfileActorTest.java
```diff
@@ -0,0 +1,135 @@
...
+  private String userId = "USER-ID";
+
+  @Before
+  public void beforeEachTest() {
+
+    PowerMockito.mockStatic(ServiceFactory.class);
+    PowerMockito.mockStatic(ElasticSearchUtil.class);
+    PowerMockito.mockStatic(Util.class);
+    searchDTO = Mockito.mock(SearchDTO.class);
+    when(Util.createSearchDto(Mockito.anyMap())).thenReturn(searchDTO);
```
amolnin: What happens if we don't mock this?
Can you check - i believe it should work without mocking - createSearchDto
if it fails, let me know - we will try to figure out solution.

PR #276, opened 2018-11-30
User skill management actor test2

kirtisagarksj:
User skill management actor test2


service/src/test/java/org/sunbird/learner/actors/skill/UserSkillManagementActorTest.java
```diff
@@ -39,39 +36,39 @@
...
-  private static FiniteDuration duration = duration("10 second");
+  private static final String ENDORSED_SKILL_NAME = "someEndorsedSkillName";
+  private FiniteDuration duration = duration("10 second");
 
   @BeforeClass
-  public static void setUp() {
-    PowerMockito.mockStatic(ElasticSearchUtil.class);
+  public static void setup() {
     system = ActorSystem.create("system");
+    PowerMockito.mockStatic(ServiceFactory.class);
```
amolnin: all static c=mocks can be added here- 
rename to setUp - codacy detects this as an issue

PR #287, opened 2018-12-03
User role actor test

kirtisagarksj:
User role actor test


amolnin:
Does not have any change for UserRole Actor test -> closing the pull request

PR #302, opened 2018-12-11
Issue #SC-593 feat : UserName Generation Method

sudhirgiri2911:
UserName Generation

### DIFF
--- a/actors/user/src/main/java/org/sunbird/user/util/UserUtil.java
+++ b/actors/user/src/main/java/org/sunbird/user/util/UserUtil.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


actors/user/src/main/java/org/sunbird/user/util/UserUtil.java
```diff
@@ -585,6 +588,43 @@ public static void toLower(Map<String, Object> userMap) {
               }
             });
   }
+
+  public static String[] userNameGenerator(String name) {
+    if (name == null || name.isEmpty()) return null;
```
bvinayakumar: As per design discussion yesterday, the decision is to generate a user name in following format.

User name (in lowercase without spaces or other special characters) concatenated with a configurable number of digits (default: 4).

sudhirgiri2911:
Closing this PR as I have mistakenly taken pull from non-analyzed branch name release-1.12.1 and have some unwanted file in my PR against new analysed branch release-1.13.

PR #303, opened 2018-12-11
Sb 2064

kirtisagarksj:
Sb 2064


actors/common/src/main/java/org/sunbird/metrics/actors/CourseMetricsActor.java
```diff
@@ -291,6 +291,7 @@ private void courseProgressMetrics(Request actorMessage) {
         map.put("user", userId);
         map.put(
             JsonKey.USERNAME, decryptionService.decryptData((String) map.get(JsonKey.USERNAME)));
+        map.put(JsonKey.USER_NAME, map.get(JsonKey.FIRST_NAME) + " " + map.get(JsonKey.LAST_NAME));
```
amolnin: check for nullability - of lastname.
Add full_name as value of concatenated full name.

PR #304, opened 2018-12-12
Issue #SC-597 feat : UserName generation

sudhirgiri2911:
UserName Generation

### DIFF
--- a/actors/user/src/main/java/org/sunbird/user/util/UserUtil.java
+++ b/actors/user/src/main/java/org/sunbird/user/util/UserUtil.java
@@ -587,31 +587,31 @@ public class UserUtil {
             });
   }
 
-  public static String[] generateUserName(String name) {
+  public static List<String> generateUsernames(String name) {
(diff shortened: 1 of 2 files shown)


actors/user/src/main/java/org/sunbird/user/util/UserUtil.java
```diff
@@ -585,6 +586,35 @@ public static void toLower(Map<String, Object> userMap) {
               }
             });
   }
+
+  public static String[] productionUserName(String name) {
+    if (name == null || name.isEmpty()) return null;
+    // configurable blocks -- number of digits to append and numberOfUserNameRequired per method
+    // call.
+    int N = 4;
```
sudhirgiri2911: Where should i put these configurable values? Should i put these in enviroment variables or external resource file.

bvinayakumar: both env and default value in resource file

sunbird_username_num_digits

PR #306, opened 2018-12-12
adding actor and util to verify key and generate otp

rahul-tarento:
adding actor and util to verify key and generate otp


actors/common/src/main/java/org/sunbird/learner/util/OTPUtil.java
```diff
@@ -0,0 +1,97 @@
...
+  private static CassandraOperation cassandraOperation = ServiceFactory.getInstance();
+  private static EncryptionService encryptionService =
+      org.sunbird.common.models.util.datasecurity.impl.ServiceFactory.getEncryptionServiceInstance(
+          null);
+  private static DbInfo userDb = Util.dbInfoMap.get(JsonKey.USER_DB);
+
+  private OTPUtil() {}
+
+  @SuppressWarnings("unchecked")
+  public static void checkPhoneUniqueness(String phone) {
```
bvinayakumar: Check phone and email uniqueness against cassandra DB. Add these methods in UserService.

actors/common/src/main/java/org/sunbird/learner/actors/otp/service/OTPService.java
```diff
@@ -0,0 +1,52 @@
...
+import org.sunbird.common.models.response.Response;
+import org.sunbird.common.models.util.JsonKey;
+import org.sunbird.common.models.util.LoggerEnum;
+import org.sunbird.common.models.util.ProjectLogger;
+import org.sunbird.common.models.util.PropertiesCache;
+import org.sunbird.helper.ServiceFactory;
+import org.sunbird.learner.util.Util;
+import org.sunbird.learner.util.Util.DbInfo;
+
+public class OTPService {
```
bvinayakumar: These methods needs to be part of OTPDaoImpl.java.

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -251,6 +251,8 @@ private static void initializeDBProperty() {
     dbInfoMap.put(
         BadgingJsonKey.USER_BADGE_ASSERTION_DB,
         getDbInfoObject(KEY_SPACE_NAME, "user_badge_assertion"));
+
+    dbInfoMap.put(JsonKey.OTP, getDbInfoObject(KEY_SPACE_NAME, "otp"));
```
bvinayakumar: Let us not add each table in dbInfoMap. Instead define below constant in corresponding DaoImpl class.

OTP_TABLE_NAME = otp

actors/common/src/main/java/org/sunbird/learner/actors/user/service/UserService.java
```diff
@@ -0,0 +1,77 @@
...
+	  private EncryptionService encryptionService =
+	      org.sunbird.common.models.util.datasecurity.impl.ServiceFactory.getEncryptionServiceInstance(
+	          null);
+	  private DbInfo userDb = Util.dbInfoMap.get(JsonKey.USER_DB);
+
+	@SuppressWarnings("unchecked")
+	  public void checkPhoneUniqueness(String phone) {
+	    // Get Phone configuration if not found , by default phone will be unique across
+	    // the application
+	    String phoneSetting = DataCacheHandler.getConfigSettings().get(JsonKey.PHONE_UNIQUE);
```
bvinayakumar: We can assume that phone and email is unique within our system for users. Therefore, we can remove the if checks on phone / email uniqueness.

actors/common/src/main/java/org/sunbird/learner/actors/user/service/UserService.java
```diff
@@ -0,0 +1,77 @@
...
+	    // Get Email configuration if not found , by default Email can be duplicate
+	    // across the
+	    // application
+	    String emailSetting = DataCacheHandler.getConfigSettings().get(JsonKey.EMAIL_UNIQUE);
+	    if (StringUtils.isNotBlank(emailSetting)
+	        && Boolean.parseBoolean(emailSetting)
+	        && StringUtils.isNotBlank(email)) {
+	      String encEmail = null;
+	      try {
+	        encEmail = encryptionService.encryptData(email);
```
bvinayakumar: We can have a helper method for checking uniqueness as the logic is similar for both phone and email uniqueness check.

actors/common/src/main/java/org/sunbird/learner/util/OTPUtil.java
```diff
@@ -0,0 +1,104 @@
...
+    emailTemplateMap.put(JsonKey.RECIPIENT_EMAILS, reciptientsMail);
+    emailTemplateMap.put(JsonKey.EMAIL_TEMPLATE_TYPE, JsonKey.OTP);
+    request = new Request();
+    request.setOperation(BackgroundOperations.emailService.name());
+    request.put(JsonKey.EMAIL_REQUEST, emailTemplateMap);
+    return request;
+  }
+
+  public static String getOTPExpirationInMinutes() {
+    String expirationInSeconds =
```
bvinayakumar: Use ProjectUtil.getConfigValue

actors/common/src/main/java/org/sunbird/learner/util/OTPUtil.java
```diff
@@ -0,0 +1,104 @@
...
+
+public final class OTPUtil {
+
+  private static final int MINIMUM_OTP_LENGTH = 6;
+  private static final int SECONDS_IN_MINUTES = 60;
+
+  private OTPUtil() {}
+
+  public static String generateOTP() {
+    String otpSize = PropertiesCache.getInstance().getProperty(JsonKey.SUNBIRD_OTP_LENGTH);
```
bvinayakumar: Use ProjectUtil.getConfigValue as it checks env variable first and then the resource properties.

PR #308, opened 2018-12-12
Sc 545

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/tac/TNCManagementActor.java
```diff
@@ -0,0 +1,48 @@
...
+import org.sunbird.common.models.util.ActorOperations;
+import org.sunbird.common.models.util.JsonKey;
+import org.sunbird.common.request.ExecutionContext;
+import org.sunbird.common.request.Request;
+
+@ActorConfig(
+  tasks = {"userTnCAccept"},
+  asyncTasks = {}
+)
+public class TNCManagementActor extends BaseActor {
```
bvinayakumar: UserTnCActor

actors/common/src/main/java/org/sunbird/learner/actors/tac/TNCManagementActor.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public void onReceive(Request request) throws Throwable {
+    String operation = request.getOperation();
+    ExecutionContext.setRequestId(request.getRequestId());
+
+    if (operation.equalsIgnoreCase("userTnCAccept")) {
```
bvinayakumar: Use constant for operation.

PR #315, opened 2018-12-13
Sb-9795 verify framework while updateUser by using specific frameworkId

iostream04:


actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -1043,6 +1047,25 @@ public static String getFrameworkId(String hashtagId) {
     } else return frameworkId;
   }
 
+  @SuppressWarnings("unchecked")
+  public static void verifyFrameworkId(String hashtagId, String frameworkId) {
+
```
bvinayakumar: Caching part is missing. We can remove getFrameworkId if it not used anywhere.

PR #318, opened 2018-12-13
Issue feat: get user by loginId,email and phone

rajatgupta0889:


actors/user/src/main/java/org/sunbird/user/actors/UserProfileReadActor.java
```diff
@@ -587,102 +593,138 @@ private void getUserDetailsByLoginId(Request actorMessage) {
...
+        (String) actorMessage.getContext().getOrDefault(JsonKey.REQUESTED_BY, "");
+    ProjectLogger.log(
+        "requested By and requested user id == "
+            + requestedById
+            + "  "
+            + (String) result.get(JsonKey.USER_ID));
+
+    try {
+      if (!(((String) result.get(JsonKey.USER_ID)).equalsIgnoreCase(requestedById))) {
+        result = removeUserPrivateField(result);
```
manzarul: We need to provide user name as well , so remove  userName key from  projectUtil :   String[] excludes (line 65)

manzarul: From System setting , in get userProfile v2 we are removing username , after reading system setting , that also need to be updated.

PR #330, opened 2018-12-14
Verify framework for userupdate removed id from mandatory fields

iostream04:
…Id added framework verification rework


actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -168,6 +168,18 @@ private void updateUser(Request actorMessage) {
   private void validateUserFrameworkData(
       Map<String, Object> userRequestMap, Map<String, Object> userDbRecord) {
     if (userRequestMap.containsKey(JsonKey.FRAMEWORK)) {
+      Map<String, Object> framework = (Map<String, Object>) userRequestMap.get(JsonKey.FRAMEWORK);
+      List<String> frameworkId;
```
bvinayakumar: frameworkId -> frameworkIdList

PR #333, opened 2018-12-17
Address management actor test new

kirtisagarksj:
Address management actor test new

### DIFF
--- a/actors/user/src/test/java/org/sunbird/user/AddressManagementActorTest.java
+++ b/actors/user/src/test/java/org/sunbird/user/AddressManagementActorTest.java
@@ -135,7 +135,6 @@ public class AddressManagementActorTest {
   }
 
   private List<Map<String, Object>> getAddressList(boolean isDelete) {
-
     List<Map<String, Object>> lst = new ArrayList<>();
     Map<String, Object> map = new HashMap<>();
     map.put(JsonKey.ADDRESS, "anyAddress");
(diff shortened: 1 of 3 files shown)


actors/user/src/test/java/org/sunbird/user/AddressManagementActorTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+  @Test
+  public void testInsertUserAddressFailureWithoutReqParams() {
+    boolean result = testScenario(UserActorOperations.INSERT_USER_ADDRESS.getValue(), false);
+    assertTrue(result);
+  }
+
+  @Test
+  public void testUpdateUserAddressFailureWithoutMandatoryFields() {
+    boolean result = testScenario(UserActorOperations.UPDATE_USER_ADDRESS.getValue(), false);
```
amolnin: why update fail for mandatory fields? does it expect whole current object state?

actors/user/src/test/java/org/sunbird/user/AddressManagementActorTest.java
```diff
@@ -0,0 +1,138 @@
...
+    return reqObj;
+  }
+
+  private List<Map<String, Object>> getAddressList() {
+
+    List<Map<String, Object>> lst = new ArrayList<>();
+    Map<String, Object> map = new HashMap<>();
+    map.put(JsonKey.ADDRESS, "anyAddress");
+    map.put(JsonKey.ID, "someUserId");
+    map.put(JsonKey.IS_DELETED, true);
```
amolnin: If is_deleted is false - other lines will be covered - which is for additional address addition or existing address updation.

actors/user/src/test/java/org/sunbird/user/AddressManagementActorTest.java
amolnin: delete address false.  & 1. containing id, 2. not containing id are 2 other test-cases

PR #335, opened 2018-12-17
changes to set phone/email verified in case of bulkupload.

amolnin:


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/UserBulkUploadBackgroundJobActor.java
```diff
@@ -212,6 +212,15 @@ private void processUser(BulkUploadProcessTask task, String organisationId, Stri
       if (null != organisation) {
         orgName = organisation.getOrgName();
       }
+      
+      if(StringUtils.isNotEmpty(user.getPhone())) {
```
bvinayakumar: Format the code changes.

actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BulkUploadBackGroundJobActor.java
```diff
@@ -63,9 +63,8 @@
  * @author Amit Kumar
  */
 @ActorConfig(
-  tasks = {},
-  asyncTasks = {"processBulkUpload"}
-)
+    tasks = {},
```
bvinayakumar: Why is this format change showing up?

PR #349, opened 2018-12-20
Issue #SC-545 add: telemertry audit event for TNC

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/tac/UserTnCActor.java
```diff
@@ -98,17 +97,21 @@ private void acceptTNC(Request request) {
       if (((String) response.get(JsonKey.RESPONSE)).equalsIgnoreCase(JsonKey.SUCCESS)) {
         syncUserDetails(userMap);
       }
+      sender().tell(response, self());
+      Map<String, Object> targetObject = null;
```
manzarul: create a private method and call that method:
move below code inside that method:
 Map<String, Object> targetObject = null;
 
        List<Map<String, Object>> correlatedObject = new ArrayList<>();
       targetObject =
           TelemetryUtil.generateTargetObject(
               (String) userMap.get(JsonKey.USER_ID),
               JsonKey.USER,
               acceptedTnC,
               lastAcceptedVersion);
       TelemetryUtil.telemetryProcessingCall(userMap, targetObject, correlatedObject);

PR #354, opened 2018-12-21
Issue #SC-543 feat: Sub org creation should fail for inactive tenants

JaikumarSRajan:


actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -1665,6 +1663,18 @@ private String getRootOrgIdFromChannel(String channel) {
     return "";
   }
 
+  private Integer getStatusFromChannel(String channel) {
+    ProjectLogger.log(
+        "OrganisationManagementActor:getStatusFromChannel: channel = " + channel,
+        LoggerEnum.INFO.name());
+    if (!StringUtils.isBlank(channel)) {
+      List<Map<String, Object>> list = getOrg(channel);
+      if (!list.isEmpty()) return (Integer) list.get(0).getOrDefault(JsonKey.STATUS, 0);
```
manzarul: From here you are returning Wrapper Object and in caller u are converting to int. there might be scenario Status key available but value as NULL , so please handle it as well.

PR #370, opened 2018-12-28
Removed the extra logs

sudhirgiri2911:
Removed the extra logs

### DIFF
--- a/actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
+++ b/actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
@@ -38,7 +38,7 @@ public final class SchedulerManager {
   /** This method will register the quartz scheduler job. */
   private void schedule() {
     ProjectLogger.log(
-        "Call to start scheduler jobs - org.sunbird.common.quartz.scheduler.SchedulerManager",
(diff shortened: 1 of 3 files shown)


actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
```diff
@@ -313,7 +305,8 @@ public Properties setUpClusterMode() throws IOException {
       configProp.put("org.quartz.dataSource.MySqlDS.password", password);
```
manzarul: add logs here that will indicate connection is established from env.

actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
```diff
@@ -38,7 +38,8 @@ private SchedulerManager() {
   /** This method will register the quartz scheduler job. */
   private void schedule() {
     ProjectLogger.log(
-        "Call to start scheduler jobs - org.sunbird.common.quartz.scheduler.SchedulerManager");
+        "Call to start scheduler jobs - org.sunbird.common.quartz.scheduler.SchedulerManager",
```
manzarul: Make logger same as above classname:methodname . applicable all places.

PR #378, opened 2019-01-03
User update framework related test cases added

iostream04:


service/src/test/java/org/sunbird/learner/actors/UserManagementActorTest.java
```diff
@@ -49,15 +62,24 @@
...
 public class UserManagementActorTest {
 
   private ActorSystem system = ActorSystem.create("system");
   private static final Props props = Props.create(UserManagementActor.class);
 
   private static Map<String, Object> reqMap;
 
+  private static final String userId = "testUserId";
+  private static UserService userService;
+  private static SystemSettingClientImpl systemSettingClient;
```
amolnin: this is no required to be declared at class level

amolnin:
split the class with extends relationship - this is adding too many methods into one single class

iostream04:
@amolnin  Please have a look now

service/src/test/java/org/sunbird/learner/actors/UserManagementActorTest.java
```diff
@@ -50,11 +51,12 @@
...
 })
-@PowerMockIgnore({"javax.management.*"})
+@PowerMockIgnore({"javax.management.*", "javax.crypto.*", "javax.net.ssl.*", "javax.security.*"})
 public class UserManagementActorTest {
 
-  private ActorSystem system = ActorSystem.create("system");
-  private static final Props props = Props.create(UserManagementActor.class);
+  public ActorSystem system = ActorSystem.create("system");
+  public static final Props props = Props.create(UserManagementActor.class);
+  public static UserService userService;
```
amolnin: nothing should be declared public static unless it is final constant variable

service/src/test/java/org/sunbird/learner/actors/UserFrameworkUpdateTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({DataCacheHandler.class, ContentStoreUtil.class, UserServiceImpl.class, Util.class})
+public class UserFrameworkUpdateTest extends UserManagementActorTest {
+
+  private static final String userId = "testUserId";
+  private static UserExternalIdentityDaoImpl userExtDao;
+
+  @Before
+  public void beforeEachTestn() {
```
amolnin: beforeTest

service/src/test/java/org/sunbird/learner/actors/UserFrameworkUpdateTest.java
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  private void mockDatacacheHandler() {
+    Map<String, List<String>> frameworkFieldsConfigMap = new HashMap<>();
+    List<String> frameworkFieldConfig =
+        Arrays.asList("id", "medium", "gradeLevel", "board", "subject");
+    List<String> frameworkFieldConfigMan = Arrays.asList("id", "medium", "gradeLevel", "board");
+    frameworkFieldsConfigMap.put(JsonKey.FIELDS, frameworkFieldConfig);
+    frameworkFieldsConfigMap.put(JsonKey.MANDATORY_FIELDS, frameworkFieldConfigMan);
+    DataCacheHandler.setFrameworkFieldsConfig(frameworkFieldsConfigMap);
```
amolnin: u don't need to use setters, just mock and return the wishful results from functions that return values.

service/src/test/java/org/sunbird/learner/actors/UserFrameworkUpdateTest.java
```diff
@@ -0,0 +1,212 @@
...
+    Mockito.when(DataCacheHandler.getFrameworkFieldsConfig()).thenReturn(frameworkFieldsConfigMap);
+
+    Map<String, List<Map<String, String>>> frameworkCategoriesMap = new HashMap<>();
+    frameworkCategoriesMap.put("medium", getListForCategoryMap("English"));
+    frameworkCategoriesMap.put("gradeLevel", getListForCategoryMap("Grade 3"));
+    frameworkCategoriesMap.put("board", getListForCategoryMap("NCERT"));
+
+    Map<String, Map<String, List<Map<String, String>>>> x = new HashMap<>();
```
amolnin: proper names - what does x mean?

service/src/test/java/org/sunbird/learner/actors/UserFrameworkUpdateTest.java
```diff
@@ -0,0 +1,212 @@
...
+import org.sunbird.learner.util.Util;
+import org.sunbird.user.dao.impl.UserExternalIdentityDaoImpl;
+import org.sunbird.user.service.impl.UserServiceImpl;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({DataCacheHandler.class, ContentStoreUtil.class, UserServiceImpl.class, Util.class})
+public class UserFrameworkUpdateTest extends UserManagementActorTest {
+
+  private static final String userId = "testUserId";
+  private static UserExternalIdentityDaoImpl userExtDao;
```
amolnin: Why mock our internal class -> u can mock cassandra operation, encryptionService for this purpose, no need to mock internal classes

service/src/test/java/org/sunbird/learner/actors/UserManagementActorTest.java
amolnin: incorrect - we should not mock user-service it is internal class - using cassandra internally

amolnin:
There seems to be problem in splitting the code:
Create one abstract class-> BaseUserActorTest (primarily keeps or defines methods/mocking common to both sub-classes)
Create one concrete class -> UserManagementActorTest (primarily adds any methods/mocking  - required for only User related test-cases)
Create another concrete class -> UserManagementActorFrameworkFieldsTest (primarly adds any method/mocking required for only Framework Field testing)


Let's discuss once you sketchout- how the class need to be divided. 

Also -> closing this request because it is showing changes other then test-classes. Please make sure you handle both before raising the new PR - No changes other than your test-case should reflect in this commit.

PR #383, opened 2019-01-06
Base actor test

kirtisagarksj:
Base actor test


actors/user/src/test/java/org/sunbird/user/UserProfileActorTest.java
```diff
@@ -1,134 +1,90 @@
...
 
-  private Response getSuccessResponse() {
-    Response response = new Response();
-    List<Map<String, Object>> resMapList = new ArrayList<>();
-    Map<String, Object> map = new HashMap<>();
-    resMapList.add(map);
-    response.put(JsonKey.RESPONSE, resMapList);
-    return response;
+  @Override
+  protected Map<String, Object> createResponseGet() {
```
amolnin: why are we over-riding the above 3 methods

amolnin:
Please extend the base with some major class like UserManagement, OrgManagement - and then re-raise the pull -request.

The case you have taken is simple. 
And also - come up with approach other than over-riding.

PR #385, opened 2019-01-08
Isssue #SC-779 add: Add location for user

rajatgupta0889:


actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -23,11 +23,7 @@
 import org.sunbird.cassandra.CassandraOperation;
```
bvinayakumar: Unit tests missing in this PR.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -373,6 +372,19 @@ private void processUserRequest(Map<String, Object> userMap, String callerId) {
     TelemetryUtil.telemetryProcessingCall(userMap, targetObject, correlatedObject);
   }
 
+  private void validateCodeAndAddLocationIds(Map<String, Object> userMap) {
+    if (userMap.containsKey(JsonKey.LOCATION_CODES)
+        && !CollectionUtils.isEmpty((List<String>) userMap.get(JsonKey.LOCATION_CODES))) {
+
+      List<String> locationIdList =
```
bvinayakumar: UserService shouldn't access LocationRequestValidator directly. Use LocationClient to get required information.

PR #393, opened 2019-01-11
 Issue SC-779 add: userLocation in read

rajatgupta0889:


amolnin:
Update existing test-case and add verification for read part.

PR #399, opened 2019-01-17
 Issue #SC-847 add: request handling to get data from externalId

rajatgupta0889:


actors/user/src/main/java/org/sunbird/user/actors/UserProfileReadActor.java
```diff
@@ -106,11 +108,43 @@ private void getUserProfile(Request actorMessage) {
...
+        ProjectCommonException.throwClientErrorException(
+            ResponseCode.mandatoryParamsMissing,
+            MessageFormat.format(
+                ResponseCode.mandatoryParamsMissing.getErrorMessage(), JsonKey.ID_TYPE));
+      } else {
+        HashMap<String, Object> map = new HashMap<>();
+        map.put(JsonKey.EXTERNAL_ID, id);
+        map.put(JsonKey.EXTERNAL_ID_PROVIDER, provider);
+        map.put(JsonKey.EXTERNAL_ID_TYPE, idType);
+        userId = UserUtil.getUserIdFromExternalId(map);
```
bvinayakumar: Use an existing method (in UserExternalIdentityDao) to find user by external ID from Cassandra (rather than Elasticsearch which involves sync delay) as this API is used in SSO user creation where we need to ensure that no two users are created with same external ID.

PR #400, opened 2019-01-17
changes for validating org external id with channel

rahul-tarento:


actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -254,6 +254,21 @@ private void createUser(Request actorMessage) {
         return;
       }
     }
+    if (userMap.containsKey(JsonKey.ORG_EXTERNAL_ID)) {
```
amolnin: Check and throw error if org_external_id and organisationid both are submitted

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -254,6 +254,21 @@ private void createUser(Request actorMessage) {
...
+      String channel = (String) userMap.get(JsonKey.CHANNEL);
+      Organisation org = organisationClient.esGetOrgByExternalId(orgExternalId, channel);
+      if (org == null) {
+        ProjectLogger.log(
+            "UserManagementActor:createUser: No organisation with orgExternalId = "
+                + orgExternalId
+                + " and channel = "
+                + channel,
+            LoggerEnum.ERROR.name());
+        ProjectCommonException.throwClientErrorException(ResponseCode.invalidOrgData);
```
bvinayakumar: Throw invalidParameterValue.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
bvinayakumar: You can use parameterMismatch error code.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -254,6 +256,41 @@ private void createUser(Request actorMessage) {
...
+      String orgId =
+          orgExternalService.getOrgIdFromOrgExternalIdAndProvider(orgExternalId, channel);
+      if (StringUtils.isBlank(orgId)) {
+        ProjectLogger.log(
+            "UserManagementActor:createUser: No organisation with orgExternalId = "
+                + orgExternalId
+                + " and channel = "
+                + channel,
+            LoggerEnum.ERROR.name());
+        throw new ProjectCommonException(
```
amolnin: use static funciton of projectcommonexception

PR #402, opened 2019-01-17
Org External Id and Provider Changes

sudhirgiri2911:
Create Org By External Id and Provider changes

### DIFF
--- a/actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
+++ b/actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
@@ -328,7 +328,7 @@ public class OrganisationManagementActor extends BaseActor {
                   JsonKey.EXTERNAL_ID));
         }
         request.put(JsonKey.EXTERNAL_ID, passedExternalId);
-        request.put(JsonKey.PROVIDER, channel);
+        request.put(JsonKey.PROVIDER, channel.toLowerCase());
       } else {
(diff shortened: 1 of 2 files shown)


actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -331,8 +328,10 @@ private void createOrg(Request actorMessage) {
                   JsonKey.EXTERNAL_ID));
         }
         request.put(JsonKey.EXTERNAL_ID, passedExternalId);
+        request.put(JsonKey.PROVIDER, channel);
```
amolnin: tolowercase

actors/common/src/main/java/org/sunbird/learner/organisation/external/identity/service/OrgExternalService.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections.MapUtils;
+import org.sunbird.cassandra.CassandraOperation;
+import org.sunbird.common.models.response.Response;
+import org.sunbird.common.models.util.JsonKey;
+import org.sunbird.helper.ServiceFactory;
+
+public class OrgExternalService {
+
+  private final String KEYSPACE_NAME = "sunbird";
```
manzarul: KEYSPACE_NAME  and table name need to be configured in one place , that will help us to Change  keyspace name due to any reason.

PR #411, opened 2019-01-29
Fixes for the issue.

amolnin:
Due to boundary condition - when no of records was exactly in multiple
of batch-size, instead of breaking out of the loop, we were querying
cassandra one more time. Which was returning null - if no records found,
same situation was not handled.
Now handling null return value as well as fixing the boundray condition.


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/BaseBulkUploadBackgroundJobActor.java
```diff
@@ -116,6 +116,12 @@ public void processBulkUpload(
       sequenceRange.put(Constants.LTE, nextSequence);
       queryMap.put(BulkUploadJsonKey.SEQUENCE_ID, sequenceRange);
       List<BulkUploadProcessTask> tasks = bulkUploadProcessTaskDao.readByPrimaryKeys(queryMap);
+      if(tasks==null) {
```
bvinayakumar: Format code changes.

PR #429, opened 2019-02-07
Course batch bg sync

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/coursebatch/CourseBatchManagementActor.java
```diff
@@ -582,8 +578,10 @@ private void validateParticipants(List<String> participants, CourseBatch courseB
 
   private void removeParticipants(
       Map<String, Boolean> removedParticipant, String batchId, String courseId) {
+    List<String> unEnrollIds = new ArrayList<>();
     removedParticipant.forEach(
         (userId, aBoolean) -> {
+          unEnrollIds.add(userId);
```
amolnin: where is this used?

unenrollIds is not gtting used

PR #430, opened 2019-02-07
SC-879 :Ability to add a user to an org based on an External Org ID 

iostream04:


actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -1590,31 +1592,13 @@ private boolean validateUsrRequest(Map<String, Object> req) {
...
-        loginId = encryptionService.encryptData((String) data.get(JsonKey.LOGIN_ID));
-      } catch (Exception e) {
-        ProjectCommonException exception =
-            new ProjectCommonException(
-                ResponseCode.userDataEncryptionError.getErrorCode(),
-                ResponseCode.userDataEncryptionError.getErrorMessage(),
-                ResponseCode.SERVER_ERROR.getResponseCode());
-        sender().tell(exception, self());
-      }
+      requestDbMap.put(JsonKey.USER_PROVIDER, data.get(JsonKey.USER_PROVIDER));
```
amolnin: put else if and check if external id available then - use this flow. else - old flow should continue

PR #447, opened 2019-02-12
Sc 869 : During organisation details syncing with ES, ensure that externalId and provider should be consistent with that of org_external_identity table

iostream04:


actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -478,17 +478,17 @@ private void createOrg(Request actorMessage) {
 
   private void createOrgExternalIdRecord(String channel, String externalId, String orgId) {
     Map<String, Object> orgExtIdRequest = new HashMap<String, Object>();
-    orgExtIdRequest.put(JsonKey.PROVIDER, channel);
-    orgExtIdRequest.put(JsonKey.EXTERNAL_ID, externalId);
+    orgExtIdRequest.put(JsonKey.PROVIDER, channel.toLowerCase());
```
manzarul: in this method there is no null check for channel and externalId, we should not assume caller will do that check, so add this check. Incase of null or empty , either logs it as info or throw error , if it's  mandatory.

actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -478,17 +478,17 @@ private void createOrg(Request actorMessage) {
...
     orgExtIdRequest.put(JsonKey.ORG_ID, orgId);
 
     cassandraOperation.insertRecord(JsonKey.SUNBIRD, JsonKey.ORG_EXT_ID_DB, orgExtIdRequest);
   }
 
   private void deleteOrgExternalIdRecord(String channel, String externalId) {
     Map<String, String> orgExtIdRequest = new HashMap<String, String>();
-    orgExtIdRequest.put(JsonKey.PROVIDER, channel);
-    orgExtIdRequest.put(JsonKey.EXTERNAL_ID, externalId);
+    orgExtIdRequest.put(JsonKey.PROVIDER, channel.toLowerCase());
```
manzarul: same as above . and we can write common method will add provider and externalId in map and return map back, if every thing ok , same code used createOrgExternalIdRecord and  deleteOrgExternalIdRecord

actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -430,7 +430,7 @@ private void createOrg(Request actorMessage) {
 
       if (StringUtils.isNotBlank(passedExternalId)) {
         String channel = (String) request.get(JsonKey.CHANNEL);
-        createOrgExternalIdRecord(channel.toLowerCase(), passedExternalId, uniqueId);
+        createOrgExternalIdRecord(channel, passedExternalId, uniqueId);
       }
       ProjectLogger.log("Org data saved into cassandra.");
```
manzarul: change this logger as per new standard define in sunbird.

actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
amolnin: StringUtils.lowerCase - would handle null situation as well.

PR #452, opened 2019-02-12
Sc 880

kirtisagarksj:
Sc 880


actors/user/src/main/java/org/sunbird/user/dao/impl/userTypeDaoImpl.java
```diff
@@ -0,0 +1,26 @@
+package org.sunbird.user.dao.impl;
+
+import org.sunbird.cassandra.CassandraOperation;
+import org.sunbird.common.models.response.Response;
+import org.sunbird.helper.ServiceFactory;
+import org.sunbird.learner.util.Util;
+import org.sunbird.user.dao.UserTypeDao;
+
+public class userTypeDaoImpl implements UserTypeDao {
```
bvinayakumar: UserTypeDaoImpl

kirtisagarksj: done the changes

PR #460, opened 2019-02-14
sc-635 : Remove GSON dependency from Sunbird Platform

iostream04:


actors/badge/src/main/java/org/sunbird/badge/util/BadgingUtil.java
```diff
@@ -165,14 +167,23 @@ public static String createBadgerUrl(Map<String, Object> map, String uri, int pl
...
     if (!StringUtils.isBlank((String) map.get(BadgingJsonKey.EVIDENCE))) {
-      json.addProperty(BadgingJsonKey.EVIDENCE, (String) map.get(BadgingJsonKey.EVIDENCE));
+      ((ObjectNode) node).put(BadgingJsonKey.EVIDENCE, (String) map.get(BadgingJsonKey.EVIDENCE));
+    }
+    ((ObjectNode) node).put(BadgingJsonKey.CREATE_NOTIFICATION, false);
+
+    try {
+      return mapper.writeValueAsString(node);
+    } catch (JsonProcessingException e) {
+      ProjectLogger.log(
```
amolnin: this should not be info level logiging - append error data, as well as exception object in log function

PR #464, opened 2019-02-15
Time converter and add fix for timestamp

rajatgupta0889:


actors/common/src/main/java/org/sunbird/metrics/actors/CourseMetricsActor.java
```diff
@@ -174,6 +175,18 @@ private void courseProgressMetricsV2(Request actorMessage) {
     sender().tell(response, self());
   }
 
+  private void formatEnrolledOn(Map<String, Object> batchMap) {
+    String timeStamp = (String) batchMap.get(JsonKey.LAST_ACCESSED_ON);
+    try {
+      SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
```
amolnin: should be in some constants file?

And are we not supposed to read this format from some setting?

actors/common/src/main/java/org/sunbird/metrics/actors/CourseMetricsActor.java
```diff
@@ -174,6 +175,18 @@ private void courseProgressMetricsV2(Request actorMessage) {
...
+      Date parsedDate = sdf.parse(timeStamp);
+      batchMap.put(JsonKey.LAST_ACCESSED_ON, ProjectUtil.formatDate(parsedDate));
+    } catch (Exception e) { // this generic but you can control another types of exception
```
amolnin: remove unnecessary comments

PR #468, opened 2019-02-18
add phone in the data

rajatgupta0889:


actors/common/src/main/java/org/sunbird/metrics/actors/CourseMetricsActor.java
```diff
@@ -910,4 +924,16 @@ private String courseConsumptionResponseGenerator(
     }
     return result;
   }
+
+  private String maskEmailOrPhone(String encryptedEmailOrPhone, String type) {
```
amolnin: move this method to common utility...

PR #472, opened 2019-02-19
 Issue #SB-7792: add code to handle page section and get data from local

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/PageManagementActor.java
```diff
@@ -551,33 +554,58 @@ private void updatePageDataCacheHandler(Response response, Map<String, Object> p
...
+    }
+  }
+
+  private Map<String, Object> searchFromES(Map<String, Object> map, String dataSource) {
+    SearchDTO searcDto = new SearchDTO();
+    searcDto.setQuery(map.get(JsonKey.QUERY) + "");
+    searcDto.setLimit((Integer) map.get(JsonKey.LIMIT));
+    searcDto.getAdditionalProperties().put(JsonKey.FILTERS, map.get(JsonKey.FILTERS));
+    String type = "";
+    if (dataSource.equalsIgnoreCase(JsonKey.BATCH)) {
```
amolnin: JsonKey.BATCH.equalsIgnoreCase(dataSource)

actors/common/src/main/java/org/sunbird/learner/actors/PageManagementActor.java
```diff
@@ -551,33 +554,58 @@ private void updatePageDataCacheHandler(Response response, Map<String, Object> p
...
+          new Mapper<Map<String, Object>, Map<String, Object>>() {
+            @Override
+            public Map<String, Object> apply(Map<String, Object> result) {
               ProjectLogger.log(
-                  "PageManagementActor:getContentData:apply: section = " + section,
-                  LoggerEnum.DEBUG.name());
+                  "PageManagementActor:getContentData:apply: result = ", LoggerEnum.INFO.name());
+              if (MapUtils.isNotEmpty(result)) {
+                section.putAll(result);
+                section.remove(JsonKey.PARAMS);
```
amolnin: section.remove(jsonkey.Params)
section.put  for Group, index 

are repeated in if/else both sections - so

PR #479, opened 2019-02-20
SC-905 : logic added for optional "roles"

iostream04:


actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -71,6 +72,7 @@
   private static InterServiceCommunication interServiceCommunication =
       InterServiceCommunicationFactory.getInstance();
   private ActorRef systemSettingActorRef = null;
+  private static final String PUBLIC_ROLE = "PUBLIC";
```
manzarul: User ProjectUtils. UserRole instead of private static final String PUBLIC_ROLE = "PUBLIC";

PR #482, opened 2019-02-20
Batch progress sync

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/LeanerStateUpdateBackGroundActor.java
```diff
@@ -149,6 +149,8 @@ private void updateCourse(Map<String, Object> temp, Map<String, Integer> content
           // TODO: for some reason, ES indexing is failing with Timestamp value. need to
           // check and
           // correct it.
+          updateDb.put(JsonKey.BATCH_ID, course.get(JsonKey.BATCH_ID));
```
amolnin: remove above comment - if fixed - seems to be fixes for same purpose

PR #495, opened 2019-02-22
Update user organisations and AssignUserRole improvements

iostream04:
@manzarul

### DIFF
--- a/actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
+++ b/actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
@@ -209,6 +209,11 @@ public class UserManagementActor extends BaseActor {
             userOrg.put(JsonKey.HASH_TAG_ID, organisation.getHashTagId());
             if (userOrg.get(JsonKey.ROLES) != null) {
               RoleService.validateRoles((List<String>) userOrg.get(JsonKey.ROLES));
+              if (!((List<String>) userOrg.get(JsonKey.ROLES))
+                  .contains(ProjectUtil.UserRole.PUBLIC.getValue())) {
+                ((List<String>) userOrg.get(JsonKey.ROLES))
+                    .add(ProjectUtil.UserRole.PUBLIC.getValue());
+              }
             } else {
               userOrg.put(JsonKey.ROLES, Arrays.asList(ProjectUtil.UserRole.PUBLIC));
             }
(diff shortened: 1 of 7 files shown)


actors/user/src/main/java/org/sunbird/user/util/UserUtil.java
```diff
@@ -626,6 +626,29 @@ public static void validateUserExternalIds(User user) {
     return organisations;
   }
 
+  @SuppressWarnings("unchecked")
+  public static List<Map<String, Object>> getAllUserOrgDetails(String userId) {
```
amolnin: don't repeat the code - 

original method - nam to getActiveuserOrgDetails
this method name is ok - make the code common

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -209,8 +209,12 @@ private void validateUserOrganisations(Request actorMessage, boolean isPrivate)
             userOrg.put(JsonKey.HASH_TAG_ID, organisation.getHashTagId());
             if (userOrg.get(JsonKey.ROLES) != null) {
               RoleService.validateRoles((List<String>) userOrg.get(JsonKey.ROLES));
+              if (!((List<String>) userOrg.get(JsonKey.ROLES))
```
amolnin: reduce complexity of code - do not fetch same variable from map again and again.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -280,16 +284,19 @@ private void createOrUpdateOrganisations(
     }
   }
 
-  private void removeOrganisations(Map<String, Object> orgDbMap, String rootOrgId) {
+  private void removeOrganisations(
```
amolnin: only pass requested-by to function - instead of actorMessage

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -280,16 +285,19 @@ private void createOrUpdateOrganisations(
...
-    List<String> userOrgTobeRemoved = new ArrayList<>();
     for (String id : ids) {
-      userOrgTobeRemoved.add((String) ((Map<String, Object>) orgDbMap.get(id)).get(JsonKey.ID));
-    }
-    if (CollectionUtils.isNotEmpty(userOrgTobeRemoved)) {
-      userOrgDao.deleteUserOrgs(userOrgTobeRemoved);
+      UserOrg userOrg = mapper.convertValue(orgDbMap.get(id), UserOrg.class);
+      userOrg.setDeleted(true);
+      userOrg.setId((String) ((Map<String, Object>) orgDbMap.get(id)).get(JsonKey.ID));
+      userOrg.setUpdatedDate(ProjectUtil.getFormattedDate());
```
amolnin: remove deleteUserorgs function from userOrgDao  - as we are not supporting only soft delete

PR #511, opened 2019-03-05
add limit in search dto for getting the completed count from all user

rajatgupta0889:


actors/common/src/main/java/org/sunbird/metrics/actors/CourseMetricsActor.java
```diff
@@ -196,13 +198,16 @@ private void formatEnrolledOn(Map<String, Object> batchMap) {
...
+  private int getCompletedCount(String batchId, int leafNodeCount, Long limit) {
     SearchDTO searchDTO = new SearchDTO();
     Map<String, Object> filter = new HashMap<>();
     filter.put(JsonKey.BATCHES + "." + JsonKey.BATCH_ID, batchId);
     filter.put(JsonKey.BATCHES + "." + JsonKey.PROGRESS, leafNodeCount);
     searchDTO.getAdditionalProperties().put(JsonKey.FILTERS, filter);
-
+    searchDTO.setOffset(0);
+    if (limit != null) {
+      searchDTO.setLimit(limit.intValue());
```
manzarul: searchDTO.setOffset(0); need to move inside if condition

PR #521, opened 2019-03-13
Issue #SB-10876 feat: Fixed the bugs for health check

sudhirgiri2911:
Health check bugs for healthy bug


actors/common/src/main/java/org/sunbird/learner/actors/health/HealthActor.java
```diff
@@ -164,7 +164,7 @@ private void checkAllComponentHealth() {
     try {
       boolean response = ElasticSearchUtil.healthCheck();
       responseList.add(ProjectUtil.createCheckResponse(JsonKey.ES_SERVICE, !response, null));
-      isallHealthy = response;
+      if (!response) isallHealthy = response;
```
amolnin: isAllHealthy = response && isAllHealthy - can be done - without if check

PR #526, opened 2019-03-22
SB-11246 : adding all the duplicate dailcode in error message

iostream04:


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/TextbookTocActor.java
```diff
@@ -590,6 +590,7 @@ private void throwInvalidDialCodeError(Set<String> invalidDialCodes) {
       List<CSVRecord> csvRecords = csvFileParser.getRecords();
       validateCSV(csvRecords);
       Set<String> dialCodes = new HashSet<>();
+      Set<String> duplicateDialCodes = new HashSet<>();
```
manzarul: change HashSet to linkedhashSet , so that we can maintain order as well.

PR #529, opened 2019-03-25
Sb 11202

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -81,6 +81,8 @@
   public static final Map<String, DbInfo> dbInfoMap = new HashMap<>();
   public static final int RECOMENDED_LIST_SIZE = 10;
   private static PropertiesCache propertiesCache = PropertiesCache.getInstance();
+  public static final int MAXIMUM_ELASTIC_LIMIT = 10000;
```
manzarul: update  MAXIMUM_ELASTIC_LIMIT  to MAXIMUM_ELASTIC_DATA_LIMIT

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -81,6 +81,8 @@
...
+  public static final int DEFAULT_ELASTIC_LIMIT = 250;
```
manzarul: update DEFAULT_ELASTIC_LIMIT  to DEFAULT_ELASTIC_DATA_LIMIT

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -81,6 +81,8 @@
...
   public static final String KEY_SPACE_NAME = "sunbird";
```
manzarul: Don't hard code key space value here , read it from common place

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -600,6 +602,11 @@ public static SearchDTO createSearchDto(Map<String, Object> searchQueryMap) {
         search.setLimit(((BigInteger) searchQueryMap.get(JsonKey.LIMIT)).intValue());
       }
     }
+    if (search.getLimit() > MAXIMUM_ELASTIC_LIMIT
```
manzarul: We need to handle both separately.
1. if limit is more than max then set it max and offset as zero 
2. if limit+offset is more than max then reduce offset to reach max only.

actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -600,6 +601,12 @@ public static SearchDTO createSearchDto(Map<String, Object> searchQueryMap) {
         search.setLimit(((BigInteger) searchQueryMap.get(JsonKey.LIMIT)).intValue());
       }
     }
+    if (search.getLimit() > DEFAULT_ELASTIC_DATA_LIMIT) {
+      search.setLimit(DEFAULT_ELASTIC_DATA_LIMIT);
+    }
+    if (search.getLimit() + search.getOffset() > DEFAULT_ELASTIC_DATA_LIMIT) {
+      search.setOffset(DEFAULT_ELASTIC_DATA_LIMIT - search.getLimit());
```
manzarul: instead of search.setOffset(DEFAULT_ELASTIC_DATA_LIMIT - search.getLimit());
  we need to reduce limit size other wise it will have come duplicate records at end.

PR #530, opened 2019-03-26
Keycloak encryption

amit-tarento:


actors/user/src/main/java/org/sunbird/user/util/UserUtil.java
```diff
@@ -359,23 +359,10 @@ public static String encryptData(String value) {
...
-      } else {
-        ProjectCommonException.throwServerErrorException(ResponseCode.userRegUnSuccessfull, null);
-      }
-    } else {
-      String response = ssoManager.updateUser(userMap);
-      if (!(!StringUtils.isBlank(response) && response.equalsIgnoreCase(JsonKey.SUCCESS))) {
-        ProjectCommonException.throwServerErrorException(
-            ResponseCode.userUpdationUnSuccessfull, null);
-      }
+  public static void updatePassword(Map<String, Object> userMap) {
```
bvinayakumar: Tests missing from this PR.

PR #535, opened 2019-03-27
Telemetry Events

anmol2302:
Capturing Device id in the request context Util.java


actors/common/src/main/java/org/sunbird/learner/util/Util.java
```diff
@@ -939,6 +939,7 @@ public static void initializeContext(Request actorMessage, String env) {
       requestContext.put(JsonKey.ENV, env);
       requestContext.put(JsonKey.REQUEST_ID, actorMessage.getRequestId());
       requestContext.put(JsonKey.REQUEST_TYPE, JsonKey.API_CALL);
+      requestContext.put(JsonKey.DEVICE_ID, actorMessage.getContext().get(JsonKey.DEVICE_ID));
```
manzarul: actorMessage.getContext() can be actorMessage.getContext() != null ? actorMessage.getContext().get(JsonKey.DEVICE_ID): "NA"

PR #568, opened 2019-04-16
Add participants if enrollment type is invite-only

rajatgupta0889:


actors/common/src/main/java/org/sunbird/learner/actors/coursebatch/CourseBatchManagementActor.java
```diff
@@ -429,6 +429,12 @@ private void getCourseBatch(Request actorMessage) {
             (String) actorMessage.getContext().get(JsonKey.BATCH_ID));
     Response response = new Response();
     if (null != result) {
+      if (JsonKey.INVITE_ONLY.equalsIgnoreCase((String) result.get(JsonKey.ENROLLMENT_TYPE))) {
```
manzarul: As we are using one-one index for each type, so discuss with @rahul-tarento for course do we still have "searchIndex"

manzarul: I discuss with Rahul and got to know there is a configuration which is handling old and new mapping.

PR #620, opened 2019-05-09
Update OrganisationManagementActor.java

niharikasingh84:


actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -1795,12 +1795,14 @@ private void validateChannel(Map<String, Object> req) {
...
             ResponseCode.CLIENT_ERROR.getResponseCode());
       }
-      int status = (int) rootOrg.get(JsonKey.STATUS);
-      if (1 != status) {
-        ProjectCommonException.throwClientErrorException(
-            ResponseCode.errorInactiveOrg,
-            ProjectUtil.formatMessage(
-                ResponseCode.errorInactiveOrg.getErrorMessage(), JsonKey.CHANNEL, channel));
+      Object status = rootOrg.get(JsonKey.STATUS);
+      if(null !=status) {
```
bvinayakumar: Can you combine the two conditions into one statement?

PR #649, opened 2019-05-31
Update TextbookTocActor.java

niharikasingh84:


actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/TextbookTocActor.java
```diff
@@ -674,8 +681,8 @@ private void throwInvalidDialCodeError(Set<String> invalidDialCodes) {
                 });
           }
           Map<String, Object> map = new HashMap<>();
-          if (JsonKey.UPDATE.equalsIgnoreCase(mode) && StringUtils.isNotBlank(record.get(id))) {
-            String identifier = record.get(id).trim();
+          if (JsonKey.UPDATE.equalsIgnoreCase(mode) && StringUtils.isNotBlank(trimMapping.get(id))) {
+            String identifier = trimMapping.get(id).trim();
```
rahul-tarento: no need of duplicate trimming, all the values in trimMapping is already trimmed.

actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/TextbookTocActor.java
```diff
@@ -640,15 +644,18 @@ private void throwInvalidDialCodeError(Set<String> invalidDialCodes) {
       StringBuilder exceptionMsgs = new StringBuilder();
       for (int i = 0; i < csvRecords.size(); i++) {
         CSVRecord record = csvRecords.get(i);
+        Map<String,String> mappingValues=record.toMap();
+        Map<String,String> trimMapping=new HashMap<>();
+        mappingValues.keySet().forEach(key -> trimMapping.put(key.trim(), mappingValues.get(key).trim()));
```
rahul-tarento: try using entrySet, also mappingValues variable is not needed

Map<String,String> trimMapping = record.toMap().entrySet().stream().collect(Collectors.toMap(e->e.getKey().trim(), e->e.getValue().trim()));

do a null check before value trimming

change the 'trimMapping' to 'trimmedRecordMap'

actors/common/src/main/java/org/sunbird/learner/actors/bulkupload/TextbookTocActor.java
```diff
@@ -601,7 +601,11 @@ private void throwInvalidDialCodeError(Set<String> invalidDialCodes) {
     }
     try (InputStreamReader reader = new InputStreamReader(bomInputStream, character); ) {
       csvFileParser = csvFileFormat.parse(reader);
-      Map<String, Integer> csvHeaders = csvFileParser.getHeaderMap();
+       Map<String, Integer> csvHeadersMap = csvFileParser.getHeaderMap();
+      //trim Headers
+      HashMap<String, Integer> csvHeaders = new HashMap<>();
+      if(MapUtils.isNotEmpty(csvHeadersMap))
+        csvHeadersMap.keySet().forEach(key -> csvHeaders.put(key.trim(), csvHeadersMap.get(key)));
```
rahul-tarento: use curly braces after if statement, try using entryset instead of keyset

PR #654, opened 2019-06-04
Update PageManagementActor.java

niharikasingh84:


actors/common/src/main/java/org/sunbird/learner/actors/PageManagementActor.java
```diff
@@ -148,6 +148,9 @@ private void getSection(Request actorMessage) {
         sender().tell(section, self());
         return;
       }
+       else {
```
maheshkumargangula: Please write test case for this scenario.

PR #659, opened 2019-06-10
Merge 2.0.1

amit-tarento:


.gitmodules
```diff
@@ -1,11 +1,12 @@
 [submodule "actors/sunbird-utils"]
 	path = actors/sunbird-utils
 	url = https://github.com/project-sunbird/sunbird-utils.git
-	branch = release-2.1.0
+	branch = release-2.0.1
```
indrajra: Revert this file. We must refer to 2.1.0 only.

PR #666, opened 2019-06-15
issue #SC-1025 all actor now will use REST client for elastic search

iostream04:


manzarul:
Merging changes releated to elasticsearch high rest util and in new ES type is not required.

PR #671, opened 2019-07-09
Issues #SB-13507 feat: removing user external id encryption logic fro…

manzarul:
user externalId is not user PII data , so no need to do encryption

### DIFF
--- a/actors/user/src/main/java/org/sunbird/user/actors/UserExternalIdManagementActor.java
+++ b/actors/user/src/main/java/org/sunbird/user/actors/UserExternalIdManagementActor.java
@@ -18,10 +18,10 @@ import org.sunbird.common.models.response.Response;
 import org.sunbird.common.models.util.JsonKey;
 import org.sunbird.common.models.util.ProjectLogger;
 import org.sunbird.common.models.util.ProjectUtil;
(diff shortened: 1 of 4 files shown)


actors/user/src/main/java/org/sunbird/user/actors/UserExternalIdManagementActor.java
```diff
@@ -185,6 +188,9 @@ private static void deleteUserExternalId(Map<String, String> map) {
     map.remove(JsonKey.ORIGINAL_EXTERNAL_ID);
     map.remove(JsonKey.ORIGINAL_ID_TYPE);
     map.remove(JsonKey.ORIGINAL_PROVIDER);
+    if (map.containsKey(JsonKey.EXTERNAL_ID)) {
```
indrajra: What if we dont do this while deleting? Given that all external id will become plain text, am not sure if this needed.
Also, did you really intend to put it inside the map or remove (all previous lines seem to be removing)?

manzarul: Decryption is not required here , becuase data will be in plain text inside DB. SO just now i remove this check.

PR #674, opened 2019-07-29
Sb 14036

amit-tarento:


actors/user/src/main/java/org/sunbird/user/actors/ResetPasswordActor.java
```diff
@@ -0,0 +1,41 @@
...
+import org.sunbird.common.request.Request;
+import org.sunbird.common.responsecode.ResponseCode;
+import org.sunbird.services.sso.SSOManager;
+import org.sunbird.services.sso.SSOServiceFactory;
+
+@ActorConfig(
+		  tasks = {"resetPassword"},
+		  asyncTasks = {}
+		)
+public class ResetPasswordActor extends BaseActor {
```
manzarul: Java doc.

actors/user/src/main/java/org/sunbird/user/actors/ResetPasswordActor.java
```diff
@@ -0,0 +1,41 @@
...
+    
+	@Override
+	public void onReceive(Request request) throws Throwable {
+		String userId = (String)request.get(JsonKey.USER_ID);
+		String password = (String)request.get(JsonKey.PASSWORD);
+		resetPassword(userId,password);	
+	}
+	
+	private void resetPassword(String userId, String password) {
+		ProjectLogger.log("ResetPasswordActor:resetPassword: method called.",LoggerEnum.INFO.name());
```
manzarul: Do userId and password basic validation before passing to keyclaok , if not valid then logs info and send 400 client error.

actors/user/src/main/java/org/sunbird/user/actors/ResetPasswordActor.java
```diff
@@ -0,0 +1,41 @@
...
+		if(ssoManager.updatePassword(userId, password)){
+			Response response = new Response();
+			response.put(JsonKey.RESPONSE, JsonKey.SUCCESS);
+		} else {
+			ProjectCommonException.throwServerErrorException(ResponseCode.invalidUserCredentials);
```
manzarul: Add logger here. , Response should be server error not invalid credentials.

actors/user/src/main/java/org/sunbird/user/actors/ResetPasswordActor.java
amit-tarento: Already did in request validator

actors/user/src/main/java/org/sunbird/user/actors/ResetPasswordActor.java
amit-tarento: done

PR #675, opened 2019-07-30
Sb 14033

amit-tarento:


actors/common/src/main/java/org/sunbird/learner/actors/otp/OTPActor.java
```diff
@@ -43,6 +44,12 @@ private void generateOTP(Request request) {
     String type = (String) request.getRequest().get(JsonKey.TYPE);
     String key = getKey(type, request);
 
+    String userId = (String) request.getRequest().get(JsonKey.USER_ID);
+    if (StringUtils.isNotBlank(userId)) {
+      key = OTPUtil.getEmailPhoneByUserId(userId, type);
+      type = getType(type);
```
indrajra: Down below at line 61, `otpService.insertOTPDetails(type, key, otp);`
Is this function guarded enough to handle empty string? If not, please remove the default from the getType() and let it throw an exception.
Similarly, getEmailPhoneByUserId() too...

amit-tarento: done

PR #689, opened 2019-08-06
SC-1259

anmol2302:
<b>Ticket Ref:</b> [SC-1259](https://project-sunbird.atlassian.net/browse/SC-1259)
1: Doc Link : [Doc](https://project-sunbird.atlassian.net/wiki/spaces/User/pages/1066762497/Ecreds+-+Phase3+and+SB+rollout#Ecreds-Phase3andSBrollout-1.POSTprivate/user/v1/certs/add(PrivateAPI,willnotrequireauthtoken))
2: After validating Request it will assure the unique Id provided , if id is unique it will allow insertion in DB, else will throw exception

### DIFF
--- a/actors/user/src/main/java/org/sunbird/user/actors/CertificateActor.java
+++ b/actors/user/src/main/java/org/sunbird/user/actors/CertificateActor.java
@@ -1,5 +1,6 @@
 package org.sunbird.user.actors;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.sunbird.actor.router.ActorConfig;
 import org.sunbird.cassandra.CassandraOperation;
 import org.sunbird.common.exception.ProjectCommonException;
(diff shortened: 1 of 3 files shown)


actors/user/src/main/java/org/sunbird/user/actors/CertificateActor.java
```diff
@@ -1,83 +1,141 @@
...
-        ProjectLogger.log(
-                "CertificateActor:getCertificate: cert id is incorrect : "
-                        + certificatedId,
-                LoggerEnum.ERROR.name());
-        throw new ProjectCommonException(
-                ResponseCode.invalidParameter.getErrorCode(),
-                ProjectUtil.formatMessage(ResponseCode.invalidParameter.getErrorMessage(),JsonKey.CERT_ID),
-                ResponseCode.CLIENT_ERROR.getResponseCode());
-      }
+    private void assureUniqueCertId(String certificatedId) {
```
manzarul: We can make this method as reusable which will accept keyspace, table and id , based on that it will return boolean flag.
isIdentityPresent can be method name.

PR #692, opened 2019-08-08
Sb 14163

manzarul:
changes for send email/sms notification on reset password.


actors/common/src/main/java/org/sunbird/learner/actors/notificationservice/EmailServiceActor.java
```diff
@@ -113,6 +113,9 @@ private void sendMail(Request actorMessage) {
     String name = "";
     if (emails.size() == 1) {
       name = StringUtils.capitalize((String) user.get(JsonKey.FIRST_NAME));
+      if (StringUtils.isBlank(name) && (String) request.get(JsonKey.FIRST_NAME) != null) {
```
amit-tarento: Use StringUtils.isNotBlank(String) request.get(JsonKey.FIRST_NAME))

PR #708, opened 2019-08-21
Issue #SB-14377 feat: sending email/sms notification after user accou…

manzarul:
* sending emal/sms notification after account merge

### DIFF
--- a/actors/user/src/main/java/org/sunbird/user/actors/CertificateActor.java
+++ b/actors/user/src/main/java/org/sunbird/user/actors/CertificateActor.java
@@ -105,7 +105,23 @@ public class CertificateActor extends UserBaseActor {
             + " -"
             + mergerId,
         LoggerEnum.INFO.name());
-    HashMap<String, Object> mergeeUserMap = getUserByIdentifier(mergeeId);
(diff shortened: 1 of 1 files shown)


actors/user/src/main/java/org/sunbird/user/actors/CertificateActor.java
```diff
@@ -23,169 +26,295 @@
...
-                    ResponseCode.CLIENT_ERROR.getResponseCode());
-        }
+  private void sendMergeNotification(String mergeeId, String mergerId) {
+    ProjectLogger.log(
+        "CertificateActor:sendMergeNotification start sending merge notification to user "
+            + mergeeId
+            + " -"
+            + mergerId,
+        LoggerEnum.INFO.name());
+    HashMap<String, Object> mergeeUserMap = getUserByIdentifier(mergeeId);
```
amit-tarento: Instead of calling cassandra two times its better to use 
Response getRecordsByIdsWithSpecifiedColumns(
      String keyspaceName, String tableName, List<String> properties, List<String> ids);
below method from CassandraOperation interface

PR #733, opened 2019-09-12
Issue #SB-14537 updating exisiting user are failing, fixed

Hari-stackroute:


actors/user/src/main/java/org/sunbird/user/actors/TenantMigrationActor.java
```diff
@@ -369,6 +375,8 @@ private void validateChannelAndGetRootOrgId(Request request) {
     userRequest.put(JsonKey.ID, request.getRequest().get(JsonKey.USER_ID));
     userRequest.put(JsonKey.CHANNEL, request.getRequest().get(JsonKey.CHANNEL));
     userRequest.put(JsonKey.ROOT_ORG_ID, request.getRequest().get(JsonKey.ROOT_ORG_ID));
+    userRequest.put(JsonKey.FLAGS_VALUE, request.getRequest().get(JsonKey.FLAGS_VALUE));
```
manzarul: 1. Are u assiging flagValue before this call.
2.  request won't have User type as TEACHER.

Hari-stackroute: 1. yes, assigned to request object
2. now value not getting from request object.

PR #745, opened 2019-09-17
Issue #SB-14537 removed boolean flag parameter from UserFlagEnum

Hari-stackroute:


actors/common/src/main/java/org/sunbird/learner/util/UserFlagUtil.java
```diff
@@ -10,19 +10,22 @@
...
+   * @param flagEnabled
    * @return
    */
-  public static int getFlagValue(String userFlagType, boolean isFlagEnabled) {
+  public static int getFlagValue(String userFlagType, boolean flagEnabled) {
     int decimalValue = 0;
+    //if phone is verified, value is set to true
     if(userFlagType.equals(UserFlagEnum.PHONE_VERIFIED.getUserFlagType()) &&
-            isFlagEnabled== UserFlagEnum.PHONE_VERIFIED.isFlagEnabled()) {
+            flagEnabled == true) {
```
manzarul: flagEnabled == true  to flagEnabled  only.

Hari-stackroute: Done.

PR #755, opened 2019-09-19
Issue#SC-1323 Async Call to update methods

anmol2302:


actors/common/src/main/java/org/sunbird/common/ShadowUserProcessor.java
```diff
@@ -35,6 +38,7 @@
     private Map<String, String> extOrgIdMap = new HashMap<>();
     private Map<String,String>channelOrgIdMap=new HashMap<>();
     private String custodianOrgId;
+    private static final int N_THREADS=10;
```
manzarul: N_THREADS reduce value 10 to 5

actors/common/src/main/java/org/sunbird/common/ShadowUserProcessor.java
```diff
@@ -129,7 +133,8 @@ private void updateStatusInUserOrg(ShadowUser shadowUser, String id) {
     }
 
     private void updateUser(ShadowUser shadowUser) {
-            List<Map<String, Object>> esUser = getUserMatchedIdentifierFromES(shadowUser);
+        ExecutorService executor = Executors.newFixedThreadPool(N_THREADS);
```
manzarul: define ExecutorService out side other wise for all 20k it has to create new pool
* check once job is completed , close executor
* this method can take Executor ref

actors/common/src/main/java/org/sunbird/common/ShadowUserProcessor.java
anmol2302: done

actors/common/src/main/java/org/sunbird/common/ShadowUserProcessor.java
anmol2302: done

PR #769, opened 2019-09-27
Issue #SB-15085 checking emailverified and phoneverified for existing…

Hari-stackroute:
… users

### DIFF
--- a/actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
+++ b/actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
@@ -679,18 +679,8 @@ public class UserManagementActor extends BaseActor {
   private Map<String, Boolean> updatedUserFlagsMap(
       Map<String, Object> userMap, Map<String, Object> userDbRecord) {
     Map<String, Boolean> userBooleanMap = new HashMap<>();
-    if(userDbRecord.get(JsonKey.EMAIL) != null && (userDbRecord.get(JsonKey.EMAIL_VERIFIED) == null
-            || (boolean)userDbRecord.get(JsonKey.EMAIL_VERIFIED))) {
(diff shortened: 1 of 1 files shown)


actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -679,6 +679,16 @@ private void setStateValidation(
   private Map<String, Boolean> updatedUserFlagsMap(
       Map<String, Object> userMap, Map<String, Object> userDbRecord) {
     Map<String, Boolean> userBooleanMap = new HashMap<>();
+    if(userDbRecord.get(JsonKey.EMAIL) != null && userDbRecord.get(JsonKey.EMAIL_VERIFIED) == null) {
```
manzarul: need to handle scenario for not null and value as true as well. same for below code

Hari-stackroute: yes taken care for existing users with true value.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -679,6 +679,18 @@ private void setStateValidation(
   private Map<String, Boolean> updatedUserFlagsMap(
       Map<String, Object> userMap, Map<String, Object> userDbRecord) {
     Map<String, Boolean> userBooleanMap = new HashMap<>();
+    if(userDbRecord.get(JsonKey.EMAIL) != null && (userDbRecord.get(JsonKey.EMAIL_VERIFIED) == null
```
manzarul: convert into a separate method

Hari-stackroute: Done

PR #777, opened 2019-10-01
SB-15106

anmol2302:
<b>Ticket Ref:</b> [SB-15106](https://project-sunbird.atlassian.net/browse/SB-15106)
 - Removed Email and Phone duplicate check from CSV file

### DIFF
--- a/actors/common/src/main/java/org/sunbird/validator/user/UserBulkMigrationRequestValidator.java
+++ b/actors/common/src/main/java/org/sunbird/validator/user/UserBulkMigrationRequestValidator.java
@@ -25,14 +25,10 @@ import java.util.regex.Pattern;
 public class UserBulkMigrationRequestValidator {
 
     private ShadowUserUpload shadowUserMigration;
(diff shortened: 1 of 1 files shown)


actors/common/src/main/java/org/sunbird/validator/user/UserBulkMigrationRequestValidator.java
```diff
@@ -111,7 +114,7 @@ private void checkEmail(String email,int index){
         if(!isEmailValid){
             errorDetails.setErrorEnum(ErrorEnum.invalid);
         }
-        if(isEmailValid) {
+        if(isEmailValid && EMAIL_DUPLICATE_CHECK) {
```
manzarul: remove this code :         if(isEmailValid && EMAIL_DUPLICATE_CHECK) { } // will never run as of now.

anmol2302: @manzarul done

PR #787, opened 2019-10-17
issue SB-15342 fix: removing all check form name of the user, it's on…

manzarul:
…ly mandatory and should not be null or empty


manzarul:
No need - so just closing this.

PR #796, opened 2019-10-22
SC-1388

anmol2302:
<b>Ticket Ref:</b> [SC-1388](https://project-sunbird.atlassian.net/browse/SC-1388)
1: Added Actor To handel the reject use case of migration(shadow user)
2: Added MigrationUtils class responsible to perform transactions in shadow_user table


actors/user/src/main/java/org/sunbird/user/actors/MigrationActor.java
```diff
@@ -0,0 +1,46 @@
...
+import org.sunbird.common.request.Request;
+import org.sunbird.common.responsecode.ResponseCode;
+import org.sunbird.user.util.MigrationUtils;
+
+
+@ActorConfig(
+        tasks = {"migrateUser","rejectMigration"},
+        asyncTasks = {}
+)
+public class MigrationActor extends BaseActor {
```
manzarul: use TenantMigrationActor - this class definition is saying : it is used for moving custodian user to tenant user.

actors/user/src/main/java/org/sunbird/user/actors/MigrationActor.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void onReceive(Request request) throws Throwable {
+        if(StringUtils.equalsIgnoreCase(ActorOperations.REJECT_MIGRATION.getValue(),request.getOperation())){
```
manzarul: either log operation name or have else condition in case of some unknow operation came. better to have else and thow unknow operation error.

actors/user/src/main/java/org/sunbird/user/actors/MigrationActor.java
```diff
@@ -0,0 +1,46 @@
...
+            rejectMigration(request);
+        }
+    }
+
+    private void rejectMigration(Request request) {
+        String userId = (String) request.getContext().get(JsonKey.USER_ID);
```
manzarul: make sure user need to pass x-auth-header and that value should be same as passed userid.

actors/user/src/main/java/org/sunbird/user/actors/MigrationActor.java
```diff
@@ -0,0 +1,46 @@
...
+        ProjectLogger.log("MigrationActor:rejectMigration: started rejecting Migration with userId:"+userId, LoggerEnum.INFO.name());
+        if (StringUtils.isBlank(userId)) {
+            ProjectCommonException.throwClientErrorException(ResponseCode.invalidUserId);
+        }
+        ShadowUser shadowUser = MigrationUtils.getRecordByUserId(userId);
+        if (null == shadowUser) {
+            ProjectCommonException.throwClientErrorException(ResponseCode.invalidUserId);
+        }
+        MigrationUtils.markUserAsRejected(shadowUser);
```
manzarul: i am thinking instead of mark user as rejected method we can introduce updateShadowUser ,

actors/user/src/main/java/org/sunbird/user/actors/MigrationActor.java
anmol2302: no longer valid

actors/user/src/main/java/org/sunbird/user/actors/MigrationActor.java
anmol2302: done

actors/user/src/main/java/org/sunbird/user/actors/MigrationActor.java
anmol2302: done

PR #800, opened 2019-10-23
SC-1380

anmol2302:
<b>Ticket Ref:</b> SC-1380
1: Will throw an exception if provided userId is not found
2: also throw an exception if provided userId is already migrated.
3: will get the record from Cassandra on provided userId and then match userExtId if found correct then the user will be migrated else will throw an error with remaining and maxAttempts.


actors/common/src/main/java/org/sunbird/common/ShadowUserProcessor.java
```diff
@@ -456,7 +457,26 @@ private void changeStatusToMultiMatch(List<ShadowUser>shadowUserList){
     }
 
 
+    /**
+     * This filtering will be needed to avaoid update of claimStatus to MULTIMATCH
```
manzarul: typo  avaoid

actors/common/src/main/java/org/sunbird/common/ShadowUserProcessor.java
```diff
@@ -456,7 +457,26 @@ private void changeStatusToMultiMatch(List<ShadowUser>shadowUserList){
...
+     * of same user while updating.
+     *
+     * @param channel
+     * @param shadowUserList
+     */
+    private List<ShadowUser> filterByChannel(String channel, List<ShadowUser> shadowUserList) {
```
manzarul: this method is giving me feel that it will return all users who belongs to provided channel , but seems it's not true.
so please correct method name.

actors/user/src/main/java/org/sunbird/user/actors/TenantMigrationActor.java
```diff
@@ -401,4 +409,73 @@ private void rejectMigration(Request request) {
...
+    String userId = (String) request.getRequest().get(JsonKey.USER_ID);
+    String extUserId = (String) request.getRequest().get(JsonKey.USER_EXT_ID);
+    ShadowUser shadowUser = MigrationUtils.getRecordByUserId(userId);
+    if(null==shadowUser){
+      ProjectCommonException.throwClientErrorException(ResponseCode.invalidUserId);
+    }
+    if(shadowUser.getClaimStatus() == ClaimStatus.CLAIMED.getValue()){
+      ProjectCommonException.throwClientErrorException(ResponseCode.unAuthorized);
+    }
+     if (StringUtils.equalsIgnoreCase(shadowUser.getUserExtId(), extUserId)) {
```
manzarul: move all logic inside if block to a method.

actors/user/src/main/java/org/sunbird/user/actors/TenantMigrationActor.java
```diff
@@ -401,4 +409,73 @@ private void rejectMigration(Request request) {
...
+      prepareMigrationRequest(request,shadowUser,userId,extUserId);
+      ProjectLogger.log("TenantMigrationActor:selfMigrate:request prepared for user migration:"+request.getRequest(),LoggerEnum.INFO.name());
+      migrateUser(request);
+      Map<String, Object> propertiesMap = new HashMap<>();
```
manzarul: instead of hashMap , if scope is within block we can use weakHashMap

actors/user/src/main/java/org/sunbird/user/util/MigrationUtils.java
```diff
@@ -44,4 +44,13 @@ public static void markUserAsRejected(ShadowUser shadowUser) {
         updateRecord(propertiesMap, shadowUser.getChannel(), shadowUser.getUserExtId());
         ProjectLogger.log("MigrationUtils:markUserAsRejected:update in cassandra  with userExtId" + shadowUser.getUserExtId(),LoggerEnum.INFO.name());
     }
+
+    public static ShadowUser getRecord(String userId,String extUserId) {
```
manzarul: is this method getting called from different places , if yes then write java doc for that , if no then make this as private and see "static" key required or not.

actors/common/src/main/java/org/sunbird/common/ShadowUserProcessor.java
anmol2302: DONE

actors/common/src/main/java/org/sunbird/common/ShadowUserProcessor.java
anmol2302: DONE

actors/user/src/main/java/org/sunbird/user/actors/TenantMigrationActor.java
anmol2302: done

actors/user/src/main/java/org/sunbird/user/util/MigrationUtils.java
anmol2302: done

actors/user/src/main/java/org/sunbird/user/actors/TenantMigrationActor.java
anmol2302: done as per requirement

PR #806, opened 2019-10-24
SC-1073 play-2.7.2 upgradation

Hari-stackroute:
akka library changes in poms


actors/common/pom.xml
```diff
@@ -44,12 +44,12 @@
 		<dependency>
 			<groupId>com.typesafe.akka</groupId>
 			<artifactId>akka-actor_2.11</artifactId>
-			<version>2.5.16</version>
+			<version>2.5.19</version>
```
manzarul: put it in one place{{<version>2.5.19</version>}}
applicable for all repeated version

actors/user/src/test/java/org/sunbird/user/UserMergeActorTest.java
```diff
@@ -77,7 +78,7 @@ public void beforeEachTest() {
     cassandraOperation = mock(CassandraOperationImpl.class);
   }
 
-  // @Ignore
+  @Ignore
```
manzarul: class is marked as ignore , so remove method level ignore.

actors/common/pom.xml
Hari-stackroute: Done

PR #820, opened 2019-11-13
Sb 15506

manzarul:
* Changes for register  license attribute during rootOrg creation and updation.
* some code cleaning for org create and update

### DIFF
--- a/actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
+++ b/actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
@@ -1165,23 +1165,6 @@ public class OrganisationManagementActor extends BaseActor {
     cassandraOperation.upsertRecord(orgDbInfo.getKeySpace(), orgDbInfo.getTableName(), addressReq);
   }
 
-  /** Checks whether the parent Organisation exists */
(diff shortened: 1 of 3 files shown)


actors/common/src/main/java/org/sunbird/learner/actors/OrganisationManagementActor.java
```diff
@@ -1275,24 +1165,6 @@ private void upsertAddress(Map<String, Object> addressReq) {
...
-    orgMap.put(JsonKey.RELATION, relation);
-    orgMap.put(JsonKey.ORG_ID_TWO, orgId);
-    orgMap.put(JsonKey.ROOT_ORG_ID, rootOrgId);
-    cassandraOperation.insertRecord(
-        orgMapDbInfo.getKeySpace(), orgMapDbInfo.getTableName(), orgMap);
-  }
-
   /** Checks whether the parent Organisation exists */
   @SuppressWarnings("unchecked")
   public Boolean isValidParent(String parentId) {
```
amit-tarento: This method is not in use now, we can delete this

manzarul: done

PR #821, opened 2019-11-13
SC-1468

anmol2302:
<b>Ticket Ref:</b> SC-1468
1: Added Feed model and interface and its impl classes


actors/user/src/main/java/org/sunbird/user/service/FeedService.java
```diff
@@ -0,0 +1,14 @@
+package org.sunbird.user.service;
+
+import org.sunbird.common.models.response.Response;
+import org.sunbird.dto.SearchDTO;
+import org.sunbird.models.user.Feed;
+
+public interface FeedService {
```
manzarul: Interfcae we are starting with I and then name , so follow same

actors/user/src/main/java/org/sunbird/user/service/FeedService.java
```diff
@@ -0,0 +1,14 @@
...
+
+
+    Response save(Feed feed);
+    Response search(SearchDTO searchDTO);
+    void delete(String id);
```
manzarul: delete should be boolean

actors/user/src/main/java/org/sunbird/user/service/FeedService.java
manzarul: add java doc for interface and method.

actors/common/src/main/java/org/sunbird/models/Category.java
```diff
@@ -0,0 +1,15 @@
+package org.sunbird.models;
+
+import org.sunbird.common.models.util.JsonKey;
+
+public enum  Category {
+
+    MIGRATION_USER(JsonKey.MIGRATION_USER_OBJECT);
```
manzarul: why enum value is MIGRATION_USER_OBJECT.
i am thinking to use "USER_EXTERNAL_ID_VALIDATION" as enum name and value as well.

actors/common/src/main/java/org/sunbird/models/user/Feed.java
```diff
@@ -0,0 +1,158 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author anmolgupta
+ * Table name : user_feed or user_notifcation
```
manzarul: no need to keep table description here.

actors/user/src/main/java/org/sunbird/user/service/FeedService.java
anmol2302: done

actors/user/src/main/java/org/sunbird/user/service/FeedService.java
anmol2302: done

anmol2302: done

actors/common/src/main/java/org/sunbird/models/user/Feed.java
anmol2302: done

PR #829, opened 2019-11-21
Sc 1470

amit-tarento:


actors/user/src/main/java/org/sunbird/user/service/impl/FeedServiceImpl.java
```diff
@@ -1,24 +1,108 @@
...
+          LoggerEnum.INFO.name());
+      response = updateUserFeed(feed);
+    } else {
+      ProjectLogger.log("FeedServiceImpl:save Calling insertUserFeed", LoggerEnum.INFO.name());
+      response = insertUserFeed(feed);
     }
+    return response;
+  }
+
+  private Response insertUserFeed(Feed feed) {
```
manzarul: insertUserFeed to saveUserFeed , becuase one place we have used saveFeed

actors/user/src/main/java/org/sunbird/user/service/impl/FeedServiceImpl.java
```diff
@@ -1,24 +1,108 @@
...
+  }
 
-    @Override
-    public boolean delete(String id) {
-        return false;
+  private Response updateUserFeed(Feed feed) {
+    ProjectLogger.log(
+        "FeedServiceImpl:updateUserFeed fetching getRecordById for feed id " + feed.getId(),
+        LoggerEnum.INFO.name());
+    Response response =
```
manzarul: create separate method to getFeedById

actors/user/src/main/java/org/sunbird/user/service/impl/FeedServiceImpl.java
```diff
@@ -1,24 +1,108 @@
...
+            usrFeedDbInfo.getKeySpace(), usrFeedDbInfo.getTableName(), feed.getId());
+    Map<String, Object> userFeed = new HashMap<>();
+    if (null != response && null != response.getResult()) {
+      List<Map<String, Object>> userFeedList =
+          (List<Map<String, Object>>) response.getResult().get(JsonKey.RESPONSE);
+      if (CollectionUtils.isNotEmpty(userFeedList)) {
+        userFeed = userFeedList.get(0);
+      }
+    }
+    String feedData = (String) userFeed.get(JsonKey.FEED_DATA);
```
manzarul: Now there is no Feed data , it's updated as data.

actors/user/src/main/java/org/sunbird/user/service/impl/FeedServiceImpl.java
```diff
@@ -1,24 +1,108 @@
...
+    Map<String, Object> feedDataMap = new HashMap<>();
+    try {
+      feedDataMap = mapper.readValue(feedData, new TypeReference<Map<String, Object>>() {});
+    } catch (Exception ex) {
+      ProjectLogger.log(
+          "FeedServiceImpl:save: Exception occurred while converting feed data from db to List",
+          ex);
     }
+    List<String> userChannel = (List<String>) feedDataMap.get(JsonKey.CHANNEL);
```
manzarul: channel key change to prospectChannels , so make sure u are align with it.
and this is optional , so make sure , if this key not present then you don't need to append it.

PR #830, opened 2019-11-25
Sc 1511

amit-tarento:


actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.sunbird.bean.ShadowUser;
+import org.sunbird.common.models.util.JsonKey;
+import org.sunbird.feed.impl.FeedFactory;
+import org.sunbird.models.user.Feed;
+
+/** this class will be used as a Util for inserting Feed in table */
+public class FeedUtil {
+  private static IFeedService feedService = FeedFactory.getInstance();
+  private static final String ORG_MIGRATION_ACTION = "OrgMigrationAction";
```
manzarul: better to create feedAction enum , so that later we can easily add new feedtype.

actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
```diff
@@ -0,0 +1,61 @@
...
+  public static final String UNREAD = "unread";
```
manzarul: add feedStatus enum that will have read and unread

actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
```diff
@@ -0,0 +1,61 @@
...
+  public static final String READ = "read";
+
+  public static void saveFeed(ShadowUser shadowUser, List<String> userIds) {
+    Map<String, Object> reqMap = new HashMap<>();
+    reqMap.put(JsonKey.USER_ID, userIds.get(0));
```
manzarul: we are using only  first element from userIds then no need of taking list. 
either take only one userId or handle list properly

actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
```diff
@@ -0,0 +1,61 @@
...
+      feedService.insert(getFeedObj(shadowUser, userIds));
+    } else {
+      Map<String, Object> data = feedList.get(index).getData();
+      List<String> channelList = (List<String>) data.get(JsonKey.PROSPECT_CHANNELS);
+      channelList.add(shadowUser.getChannel());
+      feedService.update(feedList.get(index));
+    }
+  }
+
+  private static Feed getFeedObj(ShadowUser shadowUser, List<String> userIds) {
```
manzarul: 1 .getFeedObj giving sense , we are fetching feed data  from some storage, but here we are just creating feedObject , put some related name.
2. no need to take list of userIds as we are consuming only one id.

actors/common/src/main/java/org/sunbird/feed/impl/FeedFactory.java
```diff
@@ -0,0 +1,17 @@
...
+  private static IFeedService instance;
+
+  public static IFeedService getInstance() {
+    if (instance == null) {
+      synchronized (FeedServiceImpl.class) {
+        if (instance == null) instance = new FeedServiceImpl();
+      }
+    }
+    return instance;
+    // return new FeedServiceImpl();
```
manzarul: remove commented code

actors/common/src/main/java/org/sunbird/feed/impl/FeedFactory.java
```diff
@@ -0,0 +1,17 @@
+package org.sunbird.feed.impl;
+
+import org.sunbird.feed.IFeedService;
+
+public class FeedFactory {
+  private static IFeedService instance;
+
+  public static IFeedService getInstance() {
+    if (instance == null) {
+      synchronized (FeedServiceImpl.class) {
```
manzarul: instead of  FeedServiceImpl.class put lock on FeedFactory

actors/common/src/test/java/org/sunbird/feed/FeedUtilTest.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Test
+  public void saveFeedInsertTest() {
+    List<String> userIds = new ArrayList<>();
+    userIds.add("123-456-7890");
+    try {
+      FeedUtil.saveFeed(getShadowUser(), userIds);
+    } catch (Exception e) {
+      Assert.assertTrue(false);
```
manzarul: are u expecting this test case as failure always
becuase your assert will always be false.
add atleast some more test scenario

actors/common/src/test/java/org/sunbird/feed/impl/FeedServiceImplTest.java
```diff
@@ -0,0 +1,140 @@
...
+        .thenReturn(upsertResponse);
+    PowerMockito.when(
+            cassandraOperation.deleteRecord(
+                Mockito.anyString(), Mockito.anyString(), Mockito.anyString()))
+        .thenReturn(upsertResponse);
+  }
+
+  @Test
+  public void testInsert() {
+    initCassandraForSuccess();
```
manzarul: initCassandraForSuccess() method can be called from setup() last line. in that case we don't need to call from each method.

actors/common/src/test/java/org/sunbird/feed/impl/FeedServiceImplTest.java
```diff
@@ -0,0 +1,140 @@
...
+    Response res = feedService.update(getFeed(true));
+    Assert.assertTrue(
+        ((String) res.getResult().get(JsonKey.RESPONSE)).equalsIgnoreCase(JsonKey.SUCCESS));
+  }
+
+  @Test
+  public void testDelete() {
+    initCassandraForSuccess();
+    try {
+      feedService.delete("123-456-789");
```
manzarul: delete method is not throwing any exception ,so code inside exception won't execute.
boolean response = fasle;
 feedService.delete("123-456-789");
response = true;
Assert.assertTrue(response);

actors/user/src/test/java/org/sunbird/user/UserFeedActorTest.java
```diff
@@ -0,0 +1,106 @@
...
+    search.getAdditionalProperties().put(JsonKey.FILTERS, filters);
+    Promise<Map<String, Object>> promise = Futures.promise();
+    promise.success(esResponse);
+    when(ElasticSearchHelper.getResponseFromFuture(Mockito.any())).thenReturn(esResponse);
+    PowerMockito.when(esService.search(search, ProjectUtil.EsType.userfeed.getTypeName()))
+        .thenReturn(promise.future());
+  }
+
+  @Test
+  public void getUserFeedTest() {
```
manzarul: for each method there should be at least 2 test case.

service/src/main/resources/application.conf
```diff
@@ -573,6 +573,12 @@ SunbirdMWSystem {
             nr-of-instances = 2
             dispatcher = rr-usr-dispatcher
           }
+        "/RequestRouter/*/UserFeedActor"
+          {
+            router = smallest-mailbox-pool
+            nr-of-instances = 2
```
manzarul: increase size to 5

actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  private static Feed createFeedObj(ShadowUser shadowUser, String userId) {
+    Feed feed = new Feed();
+    feed.setPriority(1);
+    feed.setCreatedBy(shadowUser.getAddedBy());
+    feed.setStatus(FeedStatus.UNREAD.getfeedStatus());
+    feed.setCategory(FeedAction.ORG_MIGRATION_ACTION.getfeedAction());
+    Map<String, Object> prospectsChannel = new HashMap<>();
```
manzarul: 1. During object creation time we can pass priority, status and category  (no need here to set again at least as of now)
2. prospectsChannel is storing only one object , so better to put capacity as double only in this case 2 , to avoid extra object creation
same can be done for list as well.

PR #838, opened 2019-12-02
Sb 16128.1

manzarul:
chnages for adding sort by key in location search , if location search have filter and type attribute and sort


manzarul:
This PR has some other file changes as well, so closing this and will raise new one.

PR #858, opened 2019-12-20
SC-1589

anmol2302:
1: user exist operation added
2: Actor Instances changed to 25 from 5

### DIFF
--- a/actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
+++ b/actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
@@ -270,7 +270,6 @@ public final class SchedulerManager {
             .build();
     String shadowUserTime =
         PropertiesCache.getInstance().getProperty("quartz_shadow_user_migration_timer");
-    shadowUserTime = "0 0/10 * 1/1 * ? *";
     ProjectLogger.log(
         "SchedulerManager:scheduleShadowUser: schedule time is : " + shadowUserTime,
         LoggerEnum.INFO.name());
--- a/actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
+++ b/actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
@@ -270,6 +270,7 @@ public final class SchedulerManager {
             .build();
     String shadowUserTime =
         PropertiesCache.getInstance().getProperty("quartz_shadow_user_migration_timer");
+    shadowUserTime = "0 0/10 * 1/1 * ? *";
     ProjectLogger.log(
         "SchedulerManager:scheduleShadowUser: schedule time is : " + shadowUserTime,
         LoggerEnum.INFO.name());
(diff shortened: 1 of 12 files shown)


actors/common/src/main/java/org/sunbird/common/quartz/scheduler/SchedulerManager.java
```diff
@@ -270,6 +270,7 @@ private void scheduleShadowUser(String identifier) {
             .build();
     String shadowUserTime =
         PropertiesCache.getInstance().getProperty("quartz_shadow_user_migration_timer");
+    shadowUserTime = "0 0/10 * 1/1 * ? *";
```
manzarul: shadowUserTime  changes is not required , removed it.

actors/user/src/main/java/org/sunbird/user/actors/UserProfileReadActor.java
```diff
@@ -884,4 +878,31 @@ private void updateTncInfo(Map<String, Object> result) {
...
+    } catch (Exception var11) {
+      ProjectCommonException exception = new ProjectCommonException(ResponseCode.userDataEncryptionError.getErrorCode(), ResponseCode.userDataEncryptionError.getErrorMessage(), ResponseCode.SERVER_ERROR.getResponseCode());
+      this.sender().tell(exception, this.self());
+    }
+    searchMap.put((String) request.get(JsonKey.KEY),encryptedValue);
+    ProjectLogger.log("UserProfileReadActor:checkUserExistence: search map prepared "+searchMap,LoggerEnum.INFO.name());
+    CompletableFuture<Response>fut=CompletableFuture.supplyAsync(() -> {
+      Response response=cassandraOperation.getRecordsByProperties(JsonKey.SUNBIRD, JsonKey.USER, searchMap);
+      List<Map<String,Object>>respList=(List)response.get(JsonKey.RESPONSE);
+      Response resp=new Response();
```
manzarul: incase user not found with provided identifier then return 404 error , same as get user by email/phone api

service/src/main/resources/application.conf
```diff
@@ -180,7 +180,7 @@ SunbirdMWSystem {
         "/RequestRouter/*/UserProfileReadActor"
           {
             router = smallest-mailbox-pool
-            nr-of-instances = 5
+            nr-of-instances = 25
```
manzarul: here total number of actor instance will be 25*5 =125 that seems too high , 
keep             nr-of-instances = 10

actors/user/src/main/java/org/sunbird/user/actors/UserProfileReadActor.java
anmol2302: @manzarul  in case user not found we are giving 200Ok but `isExists` to false. and isExists to true when user found

service/src/main/resources/application.conf
anmol2302: @manzarul we are experimenting with this API since our CPU consumption in any of our APIs is not more than 15% and memory is always between 20 to 25%, so I think we should give a try once with this change for better utilization of CPU and ofCourse throughput. if this is wrong we can anyways revert back to nr of instances to 10

actors/user/src/main/java/org/sunbird/user/actors/UserProfileReadActor.java
manzarul: for not found provide 404

service/src/main/resources/application.conf
manzarul: ok @anmol2302

actors/user/src/main/java/org/sunbird/user/actors/UserProfileReadActor.java
anmol2302: done

PR #883, opened 2020-01-03
Loadtest release 2.6.0

manzarul:


actors/user/src/main/java/org/sunbird/user/actors/UserProfileReadActor.java
```diff
@@ -855,6 +856,38 @@ private void checkUserExistence(Request request) {
...
+    SearchDTO searchDTO=new SearchDTO();
+    searchDTO.getAdditionalProperties().put(JsonKey.FILTERS,searchMap);
+    Future<Map<String, Object>> esFuture = esUtil.search(searchDTO, EsType.user.getTypeName());
+    Future<Response> userResponse = esFuture.map(new Mapper<Map<String, Object>, Response>() {
+                      @Override
+                      public Response apply(Map<String, Object> responseMap) {
+                        List<Map<String, Object>> respList = (List) responseMap.get(JsonKey.CONTENT);
+                        long size=respList.size();
+                        Response resp=new Response();
+                        resp.put(JsonKey.EXISTS, true);
```
indrajra: The elegant way would be to say `resp.put(JsonKey.EXISTS, size<=0);`

PR #895, opened 2020-01-29
Issue#SC-1662 stop updating custodianOrgId

anmol2302:
**Ticket Ref:** [SC-1662](https://project-sunbird.atlassian.net/browse/SC-1662?filter=10743)
1: Now emailUnique, PhoneUnique, custodianOrgId will not be allow updated from the system setting set API.
2: Added Test cases for same


actors/systemsettings/src/main/java/org/sunbird/systemsettings/actors/SystemSettingsActor.java
```diff
@@ -100,7 +100,9 @@ private void setSystemSetting(Request actorMessage) {
     if (JsonKey.PHONE_UNIQUE.equalsIgnoreCase(field)
             || JsonKey.EMAIL_UNIQUE.equalsIgnoreCase(field)
             || JsonKey.PHONE_UNIQUE.equalsIgnoreCase(id)
-            || JsonKey.EMAIL_UNIQUE.equalsIgnoreCase(id)) {
+            || JsonKey.EMAIL_UNIQUE.equalsIgnoreCase(id)
```
manzarul: Can we make it more generic:
Example: define all valuses inside externalproperties file with comma separated, and make list from that value.
same can be set by adopter inside env if they want to allow set or update using api

PR #908, opened 2020-03-06
Issue #SB-18059 feat: returning ClientErrorResponse if exception occu…

amit-tarento:
…rred

### DIFF
--- a/actors/common/src/main/java/org/sunbird/learner/actors/otp/OTPActor.java
+++ b/actors/common/src/main/java/org/sunbird/learner/actors/otp/OTPActor.java
@@ -27,6 +27,8 @@ public class OTPActor extends BaseActor {
 
   private OTPService otpService = new OTPService();
   private static final String SUNBIRD_OTP_ALLOWED_ATTEMPT = "sunbird_otp_allowed_attempt";
+  private static final String REMAINING_ATTEMPT = "remainingAttempt";
+  private static final String MAX_ALLOWED_ATTEMPT = "maxAllowedAttempt";
   private RateLimitService rateLimitService = new RateLimitServiceImpl();
 
   @Override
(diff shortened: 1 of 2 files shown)


actors/common/src/main/java/org/sunbird/learner/actors/otp/OTPActor.java
```diff
@@ -105,43 +106,49 @@ private void verifyOTP(Request request) {
...
-        ResponseCode.otpVerificationFailed,
-        MessageFormat.format(ResponseCode.otpVerificationFailed.getErrorMessage(), remainingCount));
+    ProjectCommonException ex = new ProjectCommonException(
+            ResponseCode.otpVerificationFailed.getErrorCode(),
+            MessageFormat.format(ResponseCode.otpVerificationFailed.getErrorMessage(), remainingCount),ResponseCode.CLIENT_ERROR.getResponseCode());
+
+    ClientErrorResponse response = new ClientErrorResponse();
+    response.setException(ex);
+    response.getResult().put("maxAllowedAttempt",Integer.parseInt(ProjectUtil.getConfigValue(SUNBIRD_OTP_ALLOWED_ATTEMPT)));
+    response.getResult().put("remainingAttempt",remainingCount);
```
anmol2302: we can use  JsonKey for `remainingAttempt`, `maxAllowedAttempt `

actors/common/src/main/java/org/sunbird/learner/actors/otp/OTPActor.java
```diff
@@ -105,43 +106,49 @@ private void verifyOTP(Request request) {
     Map<String, Object> otpDetails = otpService.getOTPDetails(type, key);
     if (MapUtils.isEmpty(otpDetails)) {
       ProjectLogger.log(
-          "OTPActor:verifyOTP: Details not found for type = " + type + " key = " + key,
-          LoggerEnum.DEBUG);
+              "OTPActor:verifyOTP: Details not found for type = " + type + " key = " + key,
+              LoggerEnum.DEBUG);
```
anmol2302: @amit-tarento can we change LoggerEnum.DEBUG TO info.

amit-tarento: Why error , its just a message , not error message

PR #912, opened 2020-04-04
Issue #SB-18620 allowing all kind of failed records to process once a…

Hari-stackroute:
…gain from the bulkupload

### DIFF
--- a/actors/common/src/main/java/org/sunbird/common/quartz/scheduler/ShadowUserMigrationScheduler.java
+++ b/actors/common/src/main/java/org/sunbird/common/quartz/scheduler/ShadowUserMigrationScheduler.java
@@ -234,7 +234,8 @@ public class ShadowUserMigrationScheduler extends BaseJob{
      * @param existingShadowUser
      */
     private void updateUserInShadowDb(String processId,MigrationUser migrationUser, ShadowUser existingShadowUser) {
-        if(!isSame(existingShadowUser,migrationUser)){
+        if((existingShadowUser.getClaimStatus() == ClaimStatus.CLAIMED.getValue() && !isSame(existingShadowUser,migrationUser)) ||
+          (!(existingShadowUser.getClaimStatus() == ClaimStatus.CLAIMED.getValue()))) {
             Map<String, Object> propertiesMap =populateUserMap(processId, migrationUser);
             if (existingShadowUser.getClaimStatus() != ClaimStatus.CLAIMED.getValue()) {
                 if (!isOrgExternalIdValid(migrationUser)) {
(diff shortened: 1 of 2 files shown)


actors/common/src/main/java/org/sunbird/common/quartz/scheduler/ShadowUserMigrationScheduler.java
```diff
@@ -231,19 +231,12 @@ private void updateUser(String processId,MigrationUser migrationUser, ShadowUser
      * ALSO this method will allow only CLAIMED user record to be updated in diksha db.
      * @param processId
      * @param migrationUser
-     * @param shadowUser
+     * @param existingShadowUser
```
indrajra, line 234: The comments above require change.

actors/common/src/main/java/org/sunbird/common/quartz/scheduler/ShadowUserMigrationScheduler.java
```diff
@@ -231,19 +231,12 @@ private void updateUser(String processId,MigrationUser migrationUser, ShadowUser
...
-            propertiesMap.put(JsonKey.EMAIL,migrationUser.getEmail());
-            propertiesMap.put(JsonKey.PHONE,migrationUser.getPhone());
-            propertiesMap.put(JsonKey.PROCESS_ID,processId);
-            propertiesMap.put(JsonKey.NAME, migrationUser.getName());
-            propertiesMap.put(JsonKey.ORG_EXT_ID, migrationUser.getOrgExternalId());
-            propertiesMap.put(JsonKey.UPDATED_ON, new Timestamp(System.currentTimeMillis()));
-            propertiesMap.put(JsonKey.USER_STATUS,getInputStatus(migrationUser.getInputStatus()));
-            if (shadowUser.getClaimStatus() != ClaimStatus.CLAIMED.getValue() && shadowUser.getClaimStatus() != ClaimStatus.REJECTED.getValue() && shadowUser.getClaimStatus() != ClaimStatus.FAILED.getValue()) {
+    private void updateUserInShadowDb(String processId,MigrationUser migrationUser, ShadowUser existingShadowUser) {
+        if(!isSame(existingShadowUser,migrationUser)){
```
indrajra, line 237: This condition will still fail - here;s the scenario.
1. User A was uploaded and A rejected the claim unknowningly.
2. Admin reuploaded.
isSame = true and therefore !isSame = false.

amit-tarento, line 237: what if user data is same for REJECTED, FAILED and ORGEXTIDMISMATCH claim status.

actors/common/src/main/java/org/sunbird/common/quartz/scheduler/ShadowUserMigrationScheduler.java
Hari-stackroute: done

actors/common/src/main/java/org/sunbird/common/quartz/scheduler/ShadowUserMigrationScheduler.java
Hari-stackroute: now checking status only for claimed-users.

Hari-stackroute: now checking status only for claimed-users.

PR #944, opened 2020-05-14
Sb 19330 adding PROSPECT_CHANNELS_IDS while creating feed

Hari-stackroute:


actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
```diff
@@ -60,11 +72,25 @@ private static Feed createFeedObj(ShadowUser shadowUser, String userId) {
...
+    prospectsChannel.put(JsonKey.PROSPECT_CHANNELS_IDS, getOrgDetails(shadowUser.getOrgExtId()));
     feed.setData(prospectsChannel);
     feed.setUserId(userId);
     return feed;
   }
 
+  private static List<Map<String, String>> getOrgDetails(String orgExtId) {
+    Map<String, Object> filters = new HashMap<>();
+    filters.put(JsonKey.EXTERNAL_ID, orgExtId);
+    Organisation org = organisationClient.esSearchOrgByFilter(filters).get(0);
```
indrajra: Since orgids and orgnames are pretty much constant, can you please add a simple hashmap of those - we can reduce calls to ES that is already loaded for the high reads presently?

indrajra (changes requested):

actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
```diff
@@ -42,9 +47,17 @@ public static Response saveFeed(ShadowUser shadowUser, String userId) {
     } else {
       Map<String, Object> data = feedList.get(index).getData();
       List<String> channelList = (List<String>) data.get(JsonKey.PROSPECT_CHANNELS);
-      if(!channelList.contains(shadowUser.getChannel())) {
+      if (!channelList.contains(shadowUser.getChannel())) {
         channelList.add(shadowUser.getChannel());
       }
+      List<Map<String, String>> orgList =
+          (List<Map<String, String>>) data.get(JsonKey.PROSPECT_CHANNELS);
```
indrajra: This must be PROSPECT_CHANNELS_IDS.

actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
```diff
@@ -42,9 +47,17 @@ public static Response saveFeed(ShadowUser shadowUser, String userId) {
...
+      orgList.forEach(
+          org -> {
+            if (!org.get("orgExtId").equals(shadowUser.getOrgExtId())) {
```
indrajra: orgExtId is no more. Please check by channel always and that's the only key.

indrajra (changes requested):

actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
Hari-stackroute: Right, taken care.

actors/common/src/main/java/org/sunbird/feed/FeedUtil.java
Hari-stackroute: yes used channel instead orgExtId

indrajra (approved):

PR #949, opened 2020-05-19
Issue #SH-135 managedby-user::update scenario - when email/phone is u…

Hari-stackroute:
…pdated, managedBy flag set to null

when email/phone is updated for a managedBy user, then managedBy filed is set to null.
Response:
`{
    "id": "api.user.update",
    "ver": "v1",
    "ts": "2020-05-19 12:07:34:772+0530",
    "params": {
        "resmsgid": null,
        "msgid": null,
        "err": null,
        "status": "success",
        "errmsg": null
    },
    "responseCode": "OK",
    "result": {
        "response": "SUCCESS",
        "link": "reset_password_link.html"
    }
}`

### DIFF
--- a/actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
+++ b/actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
@@ -160,17 +160,18 @@ public class UserManagementActor extends BaseActor {
     if (actorMessage.getContext().containsKey(JsonKey.PRIVATE)) {
       isPrivate = (boolean) actorMessage.getContext().get(JsonKey.PRIVATE);
     }
-    if (!isPrivate) {
-      if (StringUtils.isNotBlank(callerId)) {
-        userService.validateUploader(actorMessage);
-      } else {
-        userService.validateUserId(actorMessage);
-      }
-    }
     Map<String, Object> userMap = actorMessage.getRequest();
     userRequestValidator.validateUpdateUserRequest(actorMessage);
     validateUserOrganisations(actorMessage, isPrivate);
     Map<String, Object> userDbRecord = UserUtil.validateExternalIdsAndReturnActiveUser(userMap);
+    String managedById = (String) userDbRecord.get(JsonKey.MANAGED_BY);
+    if (!isPrivate) {
+      if (StringUtils.isNotBlank(callerId)) {
+        userService.validateUploader(actorMessage);
+      } else {
+        userService.validateUserId(actorMessage, managedById);
+      }
+    }
     validateUserFrameworkData(userMap, userDbRecord);
     validateUserTypeForUpdate(userMap);
     User user = mapper.convertValue(userMap, User.class);
@@ -207,7 +208,7 @@ public class UserManagementActor extends BaseActor {
     Map<String, Boolean> userBooleanMap = updatedUserFlagsMap(userMap, userDbRecord);
     int userFlagValue = userFlagsToNum(userBooleanMap);
     requestMap.put(JsonKey.FLAGS_VALUE, userFlagValue);
-    if (StringUtils.isNotEmpty((String) userDbRecord.get(JsonKey.MANAGED_BY))
+    if (StringUtils.isNotEmpty(managedById)
             && (StringUtils.isNotEmpty((String) requestMap.get(JsonKey.EMAIL)))
         || (StringUtils.isNotEmpty((String) requestMap.get(JsonKey.PHONE)))) {
       requestMap.put(JsonKey.MANAGED_BY, null);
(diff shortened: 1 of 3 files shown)


actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -963,6 +987,25 @@ private void sendEmailAndSms(Map<String, Object> userMap) {
     tellToAnother(EmailAndSmsRequest);
   }
 
+  private Response sendResetPasswordLink(Map<String, Object> userMap) {
```
indrajra: Isn't this already somewhere? If so, please remove that and unify these code.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -963,6 +987,25 @@ private void sendEmailAndSms(Map<String, Object> userMap) {
...
+      userMap.put(JsonKey.TYPE, JsonKey.PHONE);
+    }
+    passwordResetRequest.getRequest().putAll(userMap);
+    ProjectLogger.log(
+        "UserManagementActor: sendResetPasswordLink: sending reset password link to: "
+            + passwordResetRequest.getRequest().get(JsonKey.USER_ID),
+        LoggerEnum.INFO.name());
+    Response response =
+        (Response)
+            interServiceCommunication.getResponse(
```
indrajra: This is a bad idea - interServiceCommunication within the actor systems. Please get rid of it and call the actor directly.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -901,6 +916,15 @@ private void setStateValidation(
       userBooleanMap.put(
           JsonKey.STATE_VALIDATED, (boolean) userDbRecord.get(JsonKey.STATE_VALIDATED));
     }
+    // adding in release-3.0.0
```
indrajra, line 920: Without email, phone, setting these fields doesnt make sense. If there is any failure due to it, lets fix that.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
```diff
@@ -206,6 +207,12 @@ private void updateUser(Request actorMessage) {
     Map<String, Boolean> userBooleanMap = updatedUserFlagsMap(userMap, userDbRecord);
     int userFlagValue = userFlagsToNum(userBooleanMap);
     requestMap.put(JsonKey.FLAGS_VALUE, userFlagValue);
+    if (StringUtils.isNotEmpty((String) userDbRecord.get(JsonKey.MANAGED_BY))
```
indrajra: can you add a comment - to who will this apply? A regular user or managed user. 
A managed user cannot update an email or phone (UI will not allow it).
If it happens in future, then we will accept it and set managedBy = null. This managed user has become an adult and is no more dependent on the regular user.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
Hari-stackroute: I searched no-where in our code base this is present. Normally from ResetPasswordController "resetPassword" actor is called. In this use-case we need reset-password link to be generated to called from here.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
Hari-stackroute: @indrajra if we want to call a actor from another actor, we are calling tellToAnother() in the BaseActor, but whatever actor we call from this is purely async calls. Here in my scenario sync call is needed so tellToAnother() cannot be used. Even InterServiceCommunicationImpl.getResponse also calling the actor the same way you suggested. Correct me if i am wrong.

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
Hari-stackroute: @indrajra the following is suggested in the [ticket](https://project-sunbird.atlassian.net/browse/SH-135) .

- If an email or phone is added to a user that has a non-null ‘managedBy’, do the following. We shall consider these no more as managed users.
- generate a reset password link

actors/user/src/main/java/org/sunbird/user/actors/UserManagementActor.java
Hari-stackroute: @indrajra email and phone are part of it.
This is suggested in the [ticket](https://project-sunbird.atlassian.net/browse/SH-135)

PR #950, opened 2020-05-19
SH-136 Accept Terms and Conditions API changes

reshmi-nair:
- Changes to be made to take additional 'userId' param 
- accept user id(uuid) of managed profile  and mark as tnc accepted


actors/common/src/main/java/org/sunbird/learner/actors/tac/UserTnCActor.java
```diff
@@ -82,6 +88,16 @@ private void acceptTNC(Request request) {
           ResponseCode.RESOURCE_NOT_FOUND.getResponseCode());
     }
 
+    // Check whether user account is a child account(passed in request) wth managedBy is empty
```
indrajra: use managedBy only. And should this be userId or requestUserId ?

reshmi-nair: changed as per suggestion

indrajra (approved):

